PR #90, opened 2025-12-29
Add allocator to game_init hook payload

### LINKED ISSUE #89: Pass allocator in game_init hook payload
## Summary

Currently `game_init` hook payload is `void`, which limits what can be initialized in this hook.

## Problem

Component `onAdd` callbacks fire during scene loading, **before** script `init()` runs. If a component callback needs access to a subsystem (like a task engine), that subsystem must be initialized before scene loading begins.

`game_init` is the right place for this, but without access to an allocator, it's difficult to initialize allocating subsystems.

## Proposed Solution

Change `game_init` payload from `void` to include the allocator:

```zig
// In hooks/types.zig
pub const HookPayload = union(EngineHook) {
    game_init: struct { allocator: Allocator },
    // ...
};
```

## Use Case

In bakery-game, the `Worker` component has an `onAdd` callback that needs to register with `labelle-tasks`. The task engine needs to be created before any `Worker` components are added, which happens during scene loading.

With this change:

```zig
pub fn game_init(payload: engine.HookPayload) void {
    const allocator = payload.game_init.allocator;
    task_state.initTaskEngine(allocator);
}
```

apotema:
## Summary

- Add `GameInitInfo` struct with an `allocator` field to the `game_init` hook payload
- This enables subsystems to be initialized early in `game_init`, before component callbacks fire
- Export `GameInitInfo` from hooks module for user access as `engine.GameInitInfo`

Closes #89

## Motivation

When using component lifecycle callbacks (`onAdd`), components may need access to shared state that requires dynamic memory allocation. Previously, there was no way to get an allocator during `game_init`, which meant initialization had to be deferred to `scene_load` or later.

With this change, the `game_init` hook receives `GameInitInfo` with an allocator, allowing early initialization of plugins and shared state before any component callbacks fire.

## Usage

```zig
pub fn game_init(payload: engine.HookPayload) void {
    const info = payload.game_init;
    // Use info.allocator for early subsystem initialization
    my_subsystem.init(info.allocator);
}
```

## Test plan

- [x] All 335 tests pass
- [x] Updated existing tests that used `game_init = {}`
- [x] Added new test verifying allocator is accessible and usable
- [x] bakery-game builds successfully with the change

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -256,7 +256,7 @@ The engine provides a type-safe, comptime-based hook system for observing engine
 - `scene_load` / `scene_unload` - Scene transitions
 - `entity_created` / `entity_destroyed` - Entity lifecycle
 
-> **Note on `game_init`:** This hook fires during `Game.init()` before the struct is in its final memory location. Handlers should not store or use `*Game` pointers. For logic requiring a stable Game pointer, use `scene_load` or call after `game.fixPointers()`.
+> **Note on `game_init`:** This hook fires during `Game.init()` before the struct is in its final memory location. Handlers should not store or use `*Game` pointers. For logic requiring a stable Game pointer, use `scene_load` or call after `game.fixPointers()`. The `game_init` payload includes an allocator for initializing subsystems.
 
 **Basic usage:**
 ```zig
@@ -264,7 +264,10 @@ const engine = @import("labelle-engine");
 
 // Define hook handlers
 const MyHooks = struct {
-    pub fn game_init(_: engine.HookPayload) void {
+    pub fn game_init(payload: engine.HookPayload) void {
+        const info = payload.game_init;
+        // Allocator available for early initialization
+        _ = info.allocator;
         std.log.info("Game started!", .{});
     }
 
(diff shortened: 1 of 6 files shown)


apotema:
## 🤖 Cursor Agent Review

## Result

Your diff is **already applied** in the workspace:

- **`game_init` payload now includes an allocator** via `GameInitInfo` (`src/hooks/types.zig`) and `HookPayload.game_init` is no longer `void`.
- **`Game.init()` emits** `.{ .game_init = .{ .allocator = allocator } }` (`src/game.zig`).
- **`GameInitInfo` is re-exported** from `src/hooks.zig` and `src/scene.zig`.
- **Docs updated** in `CLAUDE.md` to show `payload.game_init.allocator`.
- **Tests updated** in `test/hooks_test.zig` (including an allocator usability check).
- Repo-wide search shows **no remaining** `.{ .game_init = {} }` constructions.

## Verification

I wasn’t able to execute shell commands in this environment (so I couldn’t run `zig build test` here). To verify locally, run:

```bash
zig build test
```

---
*Automated review by cursor-agent (GPT-5.2)*

PR #100, opened 2025-12-30
Fix Windows release build by using minimal CLI build config

### LINKED ISSUE #99: Windows release build fails: SDL2 SDK not found
## Description

The Windows release workflow job fails because the SDL2 SDK is not available on the GitHub Actions Windows runner.

## Error

```
Could not auto-detect SDL2 sdk configuration. Please provide .zig-cache\p\SDL2-0.0.0-...\build_config\sdl.json with the following contents filled out:
{
  "x86_64-windows-gnu": {
    "include": "<path to sdk>/include",
    "libs": "<path to sdk>/lib",
    "bin": "<path to sdk>/bin"
  }
}

You can obtain a SDL2 sdk for Windows from https://github.com/libsdl-org/SDL/releases
thread 4208 panic: Failed to get paths for SDL2: FileNotFound
```

## Root Cause

The `labelle-gfx` dependency uses SDL2 for the sokol backend. On Windows, the SDL2 SDK needs to be downloaded and configured before the build.

## Possible Solutions

1. **Add SDL2 installation step** to the Windows workflow job:
   ```yaml
   - name: Install SDL2
     run: |
       Invoke-WebRequest -Uri "https://github.com/libsdl-org/SDL/releases/download/release-2.30.0/SDL2-devel-2.30.0-mingw.zip" -OutFile SDL2.zip
       Expand-Archive SDL2.zip -DestinationPath SDL2
       # Configure sdl.json or set environment variables
   ```

2. **Build CLI without graphics dependencies** - The `labelle` CLI doesn't need SDL2/graphics, only the engine library does. Consider splitting the build so the CLI can be built independently.

3. **Skip Windows build temporarily** until a proper solution is implemented.

## Affected Workflow

`.github/workflows/release.yml` - `build-windows-x86_64` job

## Related

- Release workflow added in PR #96
- This is blocking Windows binary distribution

apotema:
## Summary

- Created `build_cli.zig` - minimal build file for CLI that doesn't load graphics dependencies
- Updated release workflow to use `build_cli.zig` for all platforms

## Problem

The main `build.zig` eagerly loads `labelle-gfx` which requires SDL2. On Windows, SDL2 SDK isn't available on GitHub Actions runners, causing the build to fail.

## Solution

The CLI only needs `zts` and `build_zon` modules - it doesn't use any graphics code. Created a separate minimal build file that only loads these dependencies.

## Test plan

- [ ] CI passes
- [ ] Merge and create a new release
- [ ] Verify all 5 platform binaries are attached (including Windows)

Fixes #99

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const build_utils = @import("build_utils.zig");
 
 /// Graphics backend selection
 pub const Backend = enum {
@@ -222,20 +223,8 @@ pub fn build(b: *std.Build) void {
     generate_step.dependOn(&run_generator.step);
 
     // Main CLI executable - unified interface for labelle projects
-    const cli_exe = b.addExecutable(.{
-        .name = "labelle",
-        .root_module = b.createModule(.{
-            .root_source_file = b.path("src/cli.zig"),
-            .target = target,
-            .optimize = optimize,
-            .imports = &.{
-                .{ .name = "zts", .module = zts },
-                .{ .name = "build_zon", .module = build_zon_mod },
-            },
-        }),
-    });
-
-    b.installArtifact(cli_exe);
+    // Uses shared build logic from build_utils.zig
+    _ = build_utils.addCli(b, target, optimize, zts, build_zon_mod);
 
     // Benchmark executable - compares ECS backend performance
     const bench_exe = b.addExecutable(.{
--- a/build_cli.zig
+++ b/build_cli.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const build_utils = @import("build_utils.zig");
 
 /// Minimal build file for the labelle CLI executable only.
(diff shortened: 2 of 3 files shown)


apotema:
## 🤖 Cursor Agent Review

### Status

Your diff is **already applied** in this workspace:

- **`.github/workflows/release.yml`** already builds with `zig build --build-file build_cli.zig ...` for all targets.
- **`build_cli.zig`** already exists and matches the contents you provided.
- **`build.zig.zon`** already includes the `zts` dependency that `build_cli.zig` uses.

### Note

I wasn’t able to run `git`/`zig` commands from here (command execution was rejected), so I verified this by directly inspecting the files. If you want to sanity-check locally, run:

```bash
zig build --build-file build_cli.zig -Doptimize=ReleaseSafe
```

---
*Automated review by cursor-agent (GPT-5.2)*

PR #251, opened 2026-01-14
ci: use labelle-cli for iOS simulator build

apotema:
## Summary
- Replace manual Zig cross-compilation with labelle-cli for iOS builds
- Remove `continue-on-error: true` since the CLI should produce successful builds

## Changes
- Install labelle-cli in CI by building from source
- Use `labelle ios build --simulator --engine-path=../..` to build with local engine
- Add `engine_version` field to CI test project's `project.labelle`
- Update `.gitignore` to exclude CI-generated directories

## How it works
The labelle-cli generates `ios/build.zig` with:
- Proper sokol framework linking (Metal, UIKit, AudioToolbox, etc.)
- `apple_a14` CPU model for NEON SIMD support (box2d physics)
- Correct iOS simulator target configuration

This should resolve the previous linking failures that occurred with manual cross-compilation.

## Test plan
- [ ] CI workflow runs successfully
- [ ] iOS simulator binary is produced

### DIFF
--- a/.github/workflows/mobile-build.yml
+++ b/.github/workflows/mobile-build.yml
@@ -86,11 +86,15 @@ jobs:
           echo "**NDK:** $ANDROID_NDK_HOME" >> $GITHUB_STEP_SUMMARY
 
   # ============================================================================
-  # iOS Build (Simulator) - Using labelle-cli
-  # The CLI generates proper build.zig with sokol framework linking
+  # iOS Build (Simulator)
+  # Uses Xcode SDK sysroot for C library headers required by box2d/sokol
+  # Uses -Dcpu=apple_a14 to enable NEON features required by box2d SIMD code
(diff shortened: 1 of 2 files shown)


apotema:
Closing this PR - the CLI approach had multiple issues:

1. **Private repo access**: labelle-cli was private (now fixed by making it public)
2. **--engine-path not fully supported**: The generate command doesn't properly handle local engine paths
3. **GitHub API issues**: InvalidResponse errors when fetching engine releases

The manual Zig cross-compilation approach that was already working (with `continue-on-error: true` for expected linking failures) is more reliable for now.

## Future work
Created issue to track CLI improvements for mobile CI support:
- Support local engine paths in generate/ios/android commands
- Better error handling for API failures
- Test CLI in CI environments

The iOS build in the original workflow verifies compilation succeeds even if framework linking fails (which is expected without full Xcode integration).

PR #308, opened 2026-02-21
fix: avoid redundant position updates for parented entities

### LINKED ISSUE #292: Child entity blinks in parent-child position inheritance example
## Description

In `usage/example_position_inheritance/`, the child entity (green square parented to a bouncing red circle) visually blinks/flickers as the parent moves.

## Steps to reproduce

```bash
cd usage/example_position_inheritance
zig build run
```

The ball bounces left-right at 300 units/sec. The square (parented at offset `(0, 80)`) should smoothly follow the ball but instead blinks.

## Investigation notes

The render pipeline sync flow is:

1. `scene.update(dt)` — script updates ball's ECS Position in-place, calls `markPositionDirty(ball)`
2. `markPositionDirtyRecursive` propagates dirty flag to child via `Children` component
3. `pipeline.sync(registry)` — iterates tracked entities, reads from ECS to compute world positions

Since `computeWorldTransform()` reads from ECS (already updated by the script before sync runs), iteration order in the `AutoArrayHashMap` should not matter — both parent and child should always get the correct position.

### Verified working correctly

- `pipeline.registry` is set in `fixPointers()` — dirty propagation to children works
- `Children` component is set on parent by `hierarchy.setParent()` during scene load
- `has_parent` flag is cached correctly on tracked entity
- The `else if (tracked.created and tracked.has_parent)` fallback (line 411) always updates parented entities even if `position_dirty` was not set
- Both dirty path (line 406) and always-update path (line 411) call `computeWorldTransform` which reads current ECS state

### Possible causes to investigate

- RetainedEngine `updatePosition` timing — does the graphics backend batch or defer updates?
- `getDeltaTime()` delegates to `retained_engine.getDeltaTime()` — check if this triggers implicit frame handling
- Double-buffering in the graphics backend that could cause a frame delay
- Raylib-specific `beginFrame()`/`endFrame()` behavior

## Environment

- Branch: `rfc/289-labelle-core`
- Backend: raylib (default)
- Scene: `scenes/hierarchy_demo.zon`
- Script: `scripts/hierarchy_test.zig`

apotema:
## Summary
- Cache last synced screen position per tracked entity in the render pipeline
- Skip `updatePosition` calls when the position hasn't changed, avoiding unnecessary spatial grid churn (remove + insert operations) in the retained engine
- The `has_parent` fallback path previously called `updatePosition` every frame for all parented entities even when neither parent nor child moved — now it checks the cached position first
- This eliminates a potential source of visual glitches at spatial grid cell boundaries

## Investigation notes

Extensive code analysis of the sync pipeline, retained engine, and spatial grid revealed:
- The ECS → pipeline sync logic is correct (reads from authoritative ECS state)
- RetainedEngine uses immediate updates with no double-buffering
- Spatial grid uses insert-first-then-remove strategy (no temporary invisibility)
- The only identified issue: redundant `updatePosition` calls causing unnecessary spatial grid operations every frame for parented entities

## Test plan
- [x] All 274 engine tests pass
- [x] `CI_TEST=1 zig build run` passes for `example_position_inheritance`
- [ ] Visual verification: run `example_position_inheritance` and confirm child entity no longer blinks

Closes #292

### DIFF
--- a/render/src/pipeline.zig
+++ b/render/src/pipeline.zig
@@ -321,10 +321,12 @@ pub const RenderPipeline = struct {
     /// Update the cached screen position and push to the retained engine.
     /// Returns true if the position was actually sent (changed from last sync).
     fn syncPosition(self: *RenderPipeline, tracked: *TrackedEntity, entity_id: EntityId, pos: GfxPosition) bool {
-        // Skip redundant updatePosition calls — avoids unnecessary spatial grid
-        // churn in the retained engine (remove-from-old-cell + insert-into-new-cell)
-        // which can cause visual glitches on cell boundaries.
-        if (pos.x == tracked.last_screen_x and pos.y == tracked.last_screen_y) {
+        // Skip redundant updatePosition calls when position hasn't meaningfully changed.
+        // Uses epsilon comparison to handle floating-point precision from hierarchy transforms.
+        const eps = 1e-2; // sub-pixel threshold
+        if (std.math.approxEqAbs(f32, pos.x, tracked.last_screen_x, eps) and
+            std.math.approxEqAbs(f32, pos.y, tracked.last_screen_y, eps))
+        {
             return false;
         }
         tracked.last_screen_x = pos.x;


apotema:
@gemini review

PR #429, opened 2026-04-13
fix(jsonc): reuse existing prefab cache in loadSceneFromSource

### LINKED ISSUE #428: loadSceneFromSource reinitializes prefab cache on every call, discarding already-loaded prefabs
## Problem

`JsoncSceneBridge.loadSceneFromSource` called `initPersistentCache()` unconditionally on every invocation, even when a prefab cache already existed on the game object. This caused already-loaded prefabs to be discarded and rebuilt from scratch every time a scene was loaded.

On Android (and other platforms where the embedded prefab directory is traversed fresh each time), this manifested as **no prefab sprites appearing** — entities were created but without any visual components because the previous cache warm-up was thrown away.

## Fix

Reuse the existing cache if one is already present:

```zig
// BEFORE — unconditionally recreates the cache
const prefab_cache = try initPersistentCache(game, prefab_dir);

// AFTER — reuse existing cache, only initialize if absent
const prefab_cache = if (game.prefab_cache_ptr) |ptr|
    @as(*PrefabCache, @ptrCast(@alignCast(ptr)))
else
    try initPersistentCache(game, prefab_dir);
```

## Impact

Any project that calls `loadSceneFromSource` more than once (e.g. scene transitions, debug scene switching) would silently lose all previously loaded prefab data on the second and subsequent calls.

apotema:
Fixes #428

## Problem

`loadSceneFromSource` called `initPersistentCache` unconditionally, discarding any prefabs previously registered via `addEmbeddedPrefab`. On Android there is no filesystem fallback in `PrefabCache.get()`, so embedded prefabs registered before `setScene` were silently lost on every scene load — all prefab entity sprites appeared missing.

## Fix

Check for an existing cache before initializing a new one:

```zig
const prefab_cache = if (game.prefab_cache_ptr) |ptr|
    @as(*PrefabCache, @ptrCast(@alignCast(ptr)))
else
    try initPersistentCache(game, prefab_dir);
```

## Verified

Confirmed on flying-platform Android: prefab sprites load correctly on first and subsequent scene loads.

### DIFF
--- a/src/jsonc_scene_bridge.zig
+++ b/src/jsonc_scene_bridge.zig
@@ -34,8 +34,12 @@ pub fn JsoncSceneBridge(comptime GameType: type, comptime Components: type) type
 
         /// Load a JSONC scene file and instantiate all entities in the ECS.
         pub fn loadScene(game: *GameType, scene_path: []const u8, prefab_dir: []const u8) !void {
-            // Load prefab cache (tries .jsonc then .zon)
-            const prefab_cache = try initPersistentCache(game, prefab_dir);
+            // Reuse the existing prefab cache when available (preserves prefabs registered
+            // via addEmbeddedPrefab before this call).
+            const prefab_cache = if (game.prefab_cache_ptr) |ptr|
+                @as(*PrefabCache, @ptrCast(@alignCast(ptr)))
+            else
+                try initPersistentCache(game, prefab_dir);
 
             try loadSceneFile(game, scene_path, prefab_cache, 0);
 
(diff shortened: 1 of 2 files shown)


apotema:
@copilot review

apotema:
@gemini review

apotema:
Superseded by #431 which landed the core fix. Filing a follow-up PR to address the remaining review comments (update cache.prefab_dir on reuse, apply same reuse pattern to loadScene).

Issue #242, opened 2026-01-13
Feature: Component references within scenes and prefabs


apotema:
## Summary

Add support for components to reference other components within the same scene or prefab. This enables common patterns like:
- A `Follow` component referencing a `Transform` on another entity
- A `Trigger` component referencing a `Door` component to open
- A `Weapon` component referencing an `Inventory` component

### Scene References
```zig
.entities = .{
    .player = .{
        .components = .{
            .Position = .{ .x = 100, .y = 100 },
            .Health = .{ .max = 100 },
        },
    },
    .health_bar = .{
        .components = .{
            .Position = .{ .x = 10, .y = 10 },
            .HealthDisplay = .{ 
                .target = .player.Health,  // Reference to player's Health component
            },
        },
    },
},
```

### Prefab References (internal)
```zig
// enemy.prefab.zon
.components = .{
    .Position = .{},
    .EnemyAI = .{},
    .Hitbox = .{ 
        .damage_source = .self.EnemyAI,  // Reference within same prefab instance
    },
},
```

### Prefab References (external/scene)
When a prefab is instantiated, it may need to reference entities in the scene:
```zig
.entities = .{
    .player = .{ ... },
    .enemy = .{
        .prefab = "enemy",
        .overrides = .{
            .EnemyAI = .{ .target = .player.Position },  // Reference scene entity
        },
    },
},
```

## Design Considerations

1. **Syntax**: How to express references in `.zon` files
   - Dot notation: `.entity_name.ComponentType`
   - Special type: `Ref(.player, .Health)`

2. **Resolution timing**: 
   - Compile-time validation where possible
   - Runtime resolution for dynamic references

3. **Lifecycle**: What happens when referenced entity/component is destroyed?
   - Null the reference?
   - Error?
   - Optional references vs required?

4. **Cross-scene references**: Should references work across scene boundaries?

5. **Serialization**: How to persist references in save files

## Related Patterns

- Unity: `[SerializeField]` with drag-drop references
- Godot: NodePath references
- Bevy: Entity references with `Query<&Transform, With<Player>>`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

apotema:
## Simplified Approach: Reference by Entity Name

Use the entity field name as a string reference:

```zig
.entities = .{
    .player = .{
        .components = .{
            .Position = .{ .x = 100, .y = 100 },
            .Health = .{ .max = 100 },
        },
    },
    .camera = .{
[...]

PR #433, opened 2026-04-15
fix(game): populate atlas display_size on sprite cache lookup

### LINKED ISSUE #240: Renderer treats atlas frame size as display size — breaks downscaled / non-1:1 atlases
## Symptom
When an atlas's PNG is downscaled relative to the original artwork (e.g. shipping a 2K texture for art that was authored at 4K), every sprite renders at half its intended on-screen area, anchored to the upper-left of where it should be. The visible quarter of the sprite is the top-left quadrant of the original.

Reproduced live on Samsung Galaxy Tab A7 (SM-T505) with `flying-platform-labelle` after halving the source PNGs from 4K to 2K and updating only the texture-space coordinates in the atlas JSON (`frame.{x,y,w,h}` and `meta.size.{w,h}`). The TexturePacker `sourceSize` / `spriteSourceSize` were left at the original values (correctly representing the design-space dimensions).

## Root cause
`labelle-gfx/src/retained_engine.zig` lines 312–318 use the atlas frame size as both the texture sub-rect AND the destination display size:

```zig
if (sprite.source_rect) |sr| {
    src_x = sr.x;
    src_y = sr.y;
    src_w = @abs(sr.width);
    src_h = @abs(sr.height);
    display_w = src_w;   // ← treats UV pixel rect as design-space size
    display_h = src_h;
}
```

`labelle-engine/src/game.zig` lines 1083–1088 sets `sprite.source_rect` from `result.sprite.getWidth()` / `getHeight()` — which return the atlas \`frame.w/h\`, i.e., texture pixel coordinates — but never threads `getSourceWidth()` / `getSourceHeight()` (the TexturePacker `sourceSize`, i.e. design dimensions) through to the renderer.

The bug stays hidden as long as the artist authors at the same resolution as the atlas (frame size == sourceSize), because all three values coincide. It only surfaces when somebody downscales the texture but keeps the design intent.

## Why this matters now
This is blocking labelle-toolkit/flying-platform-labelle#187 (Android cold start ~25s on tablet). The cold start is *entirely* PNG decode of five 4K RGBA atlases (~256 MB uncompressed). The cleanest immediate fix is to ship 2K versions of the same artwork — same on-screen quality at our 1024×768 design rect, ~6s cold start instead of 25s. But this renderer bug makes that impossible without also rewriting every prefab to use absolute design-space dimensions.

## Proposed fix
Two pieces:

1. **Atlas → engine**: in `game.zig` sprite-cache lookup, pass the atlas SpriteData's `getSourceWidth()`/`getSourceHeight()` through to the gfx layer. Either via two new fields on `SourceRect` or two new fields on the Sprite component (e.g. `intrinsic_width` / `intrinsic_height`).

2. **Renderer**: in `retained_engine.zig`, if the sprite has an explicit display size (from prefab) or a non-zero intrinsic size from the atlas, use that for `display_w/display_h` and only use `src_w/src_h` for the texture sub-rect.

Tests should cover three cases:
- Frame size == sourceSize (current default — must not regress).
- Frame size != sourceSize, no explicit prefab dims (atlas downscaled, design intent preserved).
- Frame size != sourceSize, explicit prefab dims (game wants a custom display size).

## Workaround until fixed
None. Atlas resize requires this fix or a full re-pack from source artwork at the new resolution.

apotema:
## Summary
Sprite cache lookup now passes the atlas \`SpriteData.getSourceWidth/Height()\` (TexturePacker \`sourceSize\`) into the new \`SourceRect.display_width/display_height\` fields added in labelle-toolkit/labelle-gfx#242. The renderer uses those for the destination rect when set, so sprites stay the same on-screen size even when the atlas PNG has been downscaled relative to the source artwork.

For 1:1 atlases (where artwork is authored at the same resolution as the texture, the common case) \`getSourceWidth\` returns \`getWidth\` so \`display_width == width\` and the renderer falls back to its legacy behavior. **No behavior change for existing games.**

Closes labelle-toolkit/labelle-gfx#240.

## Dependency
Requires labelle-toolkit/labelle-gfx#242 (the matching \`display_width\`/\`display_height\` fields on \`SourceRect\`). Bumping the engine pin in a project without also bumping gfx will fail to compile because the struct literal references unknown fields.

## Test plan
- [x] \`zig build\` clean (engine doesn't directly depend on gfx, so the struct literal isn't validated here)
- [x] \`zig build test\` green
- [ ] End-to-end: re-apply the 2K atlas resize on flying-platform-labelle (with this PR + #242 both wired in via local: paths) and confirm cold start drops 24s → ~8s while sprites render at the correct on-screen size

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -1067,8 +1067,18 @@ pub fn GameConfig(
                         sprite.source_rect = .{
                             .x = @floatFromInt(result.sprite.x),
                             .y = @floatFromInt(result.sprite.y),
+                            // `width`/`height` are texture-pixel coords used
+                            // for UV sampling. `display_*` carry the original
+                            // artwork's design-space size (TexturePacker
+                            // `sourceSize`) — when the atlas PNG has been
+                            // downscaled relative to the source artwork they
+                            // diverge, and the renderer uses `display_*` for
+                            // the destination rect so the on-screen sprite
+                            // size stays the same.
                             .width = @floatFromInt(result.sprite.getWidth()),
                             .height = @floatFromInt(result.sprite.getHeight()),
+                            .display_width = @floatFromInt(result.sprite.getSourceWidth()),
+                            .display_height = @floatFromInt(result.sprite.getSourceHeight()),
                         };
                         self.renderer.markVisualDirty(entity);
                     }


apotema:
Closing — same wrong shape as labelle-toolkit/labelle-gfx#242. \`getSourceWidth/Height()\` returns TexturePacker's \`sourceSize\` (the un-trimmed bounding box), which equals the display size for un-trimmed sprites but is much bigger than the rendered area for trimmed sprites. Populating \`SourceRect.display_width/height\` from it stretches every trimmed character to the un-trimmed area.

The right approach (now that the trim issue is understood) is per-atlas scale derived from \`meta.size\` vs actual texture dims, applied to source-rect UV coordinates while leaving the rendered display at the original \`frame.w/h\`. See labelle-toolkit/labelle-gfx#240 for the updated diagnosis.

Will reopen as a fresh PR once the loader-side scale plumbing is designed.

Issue #44, opened 2025-12-13
Support full entity definitions (shapes/sprites) in nested []const Entity fields


apotema:
## Problem

When defining nested entities inside component fields of type `[]const Entity`, the `createChildEntities` function in loader.zig only processes:
- `x`, `y` (position offset)
- `components`

It does **not** support:
- `shape` definitions
- `sprite` definitions
- `prefab` references

This means nested entities can only be data-only entities without visual representation.

## Real-World Example

A kitchen workstation with movement nodes (from simple_kitchen.zon):

```zon
.{
    .sprite = .{
        .name = "kitchen_room/kitchen_room1.png",
        .z_index = 10,
        .pivot = .bottom_left,
    },
    .components = .{
        .Workstation = .{
            .workstation_type = .kitchen,
            .process_duration = 5.0,
            .movement_nodes = .{
                .{
                    .shape = .{
                        .type = .circle,
                        .x = 39,
                        .y = 0,
                        .radius = 4,
                        .color = .{ .r = 100, .g = 100, .b = 255, .a = 150 },
                        .z_index = 60,
                    },
                    .components = .{
                        .Position = .{ .x = 39, .y = 0 },
                        .MovementNode = .{},
                    },
                },
                .{
                    .shape = .{
                        .type = .circle,
                        .x = 78,
                        .y = 0,
                        .radius = 4,
                        .color = .{ .r = 100, .g = 100, .b = 255, .a = 150 },
                        .z_index = 60,
                    },
                    .components = .{
                        .Position = .{ .x = 78, .y = 0 },
                        .MovementNode = .{},
                    },
                },
                .{
                    .shape = .{
                        .type = .circle,
                        .x = 117,
                        .y = 0,
                        .radius = 4,
                        .color = .{ .r = 100, .g = 100, .b = 255, .a = 150 },
                        .z_index = 60,
                    },
                    .components = .{
                        .Position = .{ .x = 117, .y = 0 },
                        .MovementNode = .{},
                    },
                },
            },
        },
    },
}
```

The `Workstation` component has:
```zig
movement_nodes: []const engine.Entity = &.{},
```

**Current behavior:** The shapes are ignored - movement nodes are created as invisible data-only entities with just Position and MovementNode components.

**Expected behavior:** The blue circles should be rendered at each movement node position.

## Expected Behavior

Nested entities in `[]const Entity` fields should support the same entity definition format as top-level scene entities:
- `.shape = .{...}` for shape entities
[...]

PR #437, opened 2026-04-15
docs(rfc): asset streaming — async decode + per-scene manifests + refcount

apotema:
## Summary

Drafts the roadmap from the current atlas-only synchronous loading to a generic worker-threaded asset catalog. Motivated by the 25 s → 11 s cold-start work this session — 11 s is still dominated by main-thread PNG decode, and the loading-screen attempt failed because the very call meant to let it animate was the call blocking the thread.

Covers all five primitives the big-engine playbook (Unity Addressables, Unreal soft refs + StreamableManager, Godot threaded ResourceLoader) converges on:

1. **Async decode** on a worker thread (decode off-thread, GPU upload back on main, since both sokol_gfx and raylib's GL backend are single-threaded by spec).
2. **Per-scene asset manifests** in scene `.jsonc` so the engine acquires/releases on transition.
3. **Reference-counted unload** so assets free automatically.
4. **Generic across asset types** — one `AssetCatalog` + per-type `AssetLoader` so atlases, audio, fonts and raw bytes share one pipeline.
5. **Backwards-compatible** — `loadAtlasFromMemory` / `loadAtlasIfNeeded` survive as sync wrappers over `acquire` + busy-wait.

Four-phase migration: phase 1 lands the primitives behind a sync shim (no caller break), phase 2 wires the scene `assets:` block and delivers the cold-start win, phase 3 adds refcount unload, phase 4 brings real audio/font loaders.

## Test plan

- [ ] Review the architecture diagram + per-layer changes against the existing engine #434 (lazy register/load split) and assembler #44 (per-resource lazy flag) that this builds on
- [ ] Confirm the open questions section captures everything that needs a decision before phase 1 starts

### DIFF
--- a/RFC-ASSET-STREAMING.md
+++ b/RFC-ASSET-STREAMING.md
@@ -1,6 +1,6 @@
 # RFC: Asset Streaming
 
-**Status:** Draft
+**Status:** Draft (revision 2 — addresses #437 review comments)
 **Author:** Alexandre
 **Date:** 2026-04-15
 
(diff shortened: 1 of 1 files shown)


apotema:
@copilot review

PR #449, opened 2026-04-15
feat(assets): AssetCatalog core + src/assets/ skeleton

apotema:
## Summary

Phase 1 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437). Lays down `src/assets/` and fully implements `AssetCatalog` so the worker (#439), real loaders (#440 / #441), pump body (#442), legacy atlas shim (#443) and scene-transition wiring (#444 / #445 / #446) can all build on top of a stable surface.

Closes #438.

## What's in

- **`src/assets/catalog.zig`** — fully implemented public `AssetCatalog` API:
  - `register` / `acquire` / `release` / `isReady` / `allReady` / `progress` / `lastError` / `pump`.
  - `StringHashMap(AssetEntry)` keyed on the plain asset name; loader is selected via `LoaderKind` at registration time and stored on the entry (no `atlas:` prefix).
  - Refcount semantics: bump on acquire, drop on release; on zero-from-`.ready` the CPU-side payload is cleared and state rewinds to `.registered`.
  - Two non-negotiable invariants documented in the file header:
    1. `AssetEntry` is owned exclusively by the main thread; transitions happen only inside `pump()` and the public methods. No mutex on the catalog — the future worker (#439) talks through bounded SPSC rings.
    2. `raw_bytes` and `file_type` are borrowed slices that live for the program lifetime via `@embedFile`; the catalog never copies them.
- **`src/assets/loader.zig`** — types only: `LoaderKind`, `DecodedPayload` union, `AssetLoaderVTable { decode, upload, drop, free }`.
- **`src/assets/worker.zig`** — types only: `WorkRequest`, `WorkResult`. Comment notes that #439 adds the `std.Thread` worker plus the SPSC rings.
- **`src/assets/loaders/{image,audio,font}.zig`** — placeholder vtables whose `decode` and `upload` return `error.NotImplemented`; `drop` and `free` are no-ops. Real loaders land in #440 / #441.
- **`src/assets/mod.zig`** — single-import re-export of the public types.
- **`src/root.zig`** — exposes `assets_mod` plus convenience aliases (`AssetCatalog`, `AssetEntry`, `AssetState`, `LoaderKind`, `DecodedPayload`, `AssetLoaderVTable`, `AssetWorkRequest`, `AssetWorkResult`).
- **`build.zig`** — adds `test/asset_catalog_test.zig` to the test list.
[...]

Issue #444, opened 2026-04-15
feat(scene): scene_assets_acquire/release hooks + setScene wiring


apotema:
Part of Phase 2 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Add two new hooks fired from `setScene` and wire the acquire-new-then-release-old transition.

- `scene_assets_acquire(target)` — fired *before* `scene_before_load` for the new scene
- `scene_assets_release(previous)` — fired *after* `scene_load` for the new scene completes

## Why acquire-new-first
Shared assets must keep refcount ≥ 1 across the swap. Releasing old first would free-then-reload any asset the new scene also needs. Standard refcount idiom.

## Transition flow
```
setScene(target):
  for asset in target.manifest.assets:
    catalog.acquire(asset)
  if not catalog.allReady(target.manifest.assets):
    setSceneAtomic(loading_scene)
    return                     # loading scene polls allReady + re-enters setScene
  setSceneAtomic(target)
  for asset in previous.manifest.assets:
    catalog.release(asset)
```

## Acceptance
- [ ] Hooks `scene_assets_acquire` / `scene_assets_release` added to the scene loader hook enum
- [ ] Default handlers implement the flow above in `scene_mixin.zig`
- [ ] Loading-scene fallback: `setSceneAtomic(loading_scene)` when anything is still decoding; loading scene's tick polls `allReady` and re-enters `setScene`
- [ ] `Game.asset_failure_policy` enum (`fatal` | `warn` | `silent`), default `fatal`, controls behavior on `.failed` asset during transition
- [ ] Tests: shared asset survives transition (refcount ≥ 1), loading-scene fallback fires, failure policy honored
- [ ] Depends on the `SceneEntry.assets` field ticket

PR #451, opened 2026-04-16
feat(scene): SceneEntry.assets field wired from assembler manifest

### LINKED ISSUE #445: feat(scene): SceneEntry.assets field + comptime manifest map
Part of Phase 2 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Add an `assets` field to `SceneEntry` and populate it from the comptime per-scene manifest map emitted by the assembler.

## Acceptance
- [ ] `SceneEntry.assets: []const []const u8` added
- [ ] Engine reads the assembler-emitted `scene_name → [asset name]` comptime map at `Game.init` and populates each `SceneEntry`
- [ ] Empty slice for scenes that omit the `assets:` block (legacy manual-load path unchanged)
- [ ] Exposed to scripts via `game.scenes.get(name).assets` per RFC example
- [ ] Depends on the labelle-assembler `assets:` parsing ticket

apotema:
## Summary

Phase 2 of the Asset Streaming RFC (#437). Plumbs each scene's parsed `"assets": [...]` manifest into `SceneEntry.assets` so runtime code can do `game.scenes.get("main").?.assets` without touching generated code.

Closes #445.

- `SceneEntry` gains `assets: []const []const u8 = &.{}`. The empty default keeps the legacy `registerSceneSimple` path working with zero churn (scripts can `.assets.len == 0` to branch on non-streaming scenes without a null check).
- `registerSceneWithAssets(name, loader, assets)` — manifest-aware overload the assembler can call directly when it wants to avoid a second map lookup. Equivalent to `registerSceneSimple` plus `setSceneAssets`.
- `setSceneAssets(name, assets) error{SceneNotFound}!void` — the setter-after-registration helper.

## API choice

**Option A (add overload) + setter-after-registration.** Both new methods live alongside the untouched `registerSceneSimple`, so every existing call site keeps compiling with zero changes. The assembler emits a single `inline for (SceneAssetManifests.entries)` loop at the end of its existing `registerSceneSimple` block — one extra comptime loop per setup-code path, not per-scene churn.

## Assembler touch

Yes, minimum required. See **labelle-toolkit/labelle-assembler#52** — that PR adds the two-line comptime `inline for` loop in `buildSetupCode` and `buildCallbackInitCode`. **It is a required merge dependency:** this engine PR exposes `setSceneAssets`, and the assembler PR is the only caller.

## Back-compat

- Legacy `registerSceneSimple` scenes see `SceneEntry.assets == &.{}` (empty slice, not null) — verified by the new `SceneEntry.assets defaults to empty for legacy registration` test.
- Slash-named scenes like `"world/intro"` stay retrievable via their original name — the flatten-to-ident step only affects the `SceneAssetManifests.<ident>` decls, not `Entry.name` or the map key. Verified by the new `slash-named scene preserves original name for lookup` test.
- The new `SceneEntry.assets` field has a default, so any user code that builds a `SceneEntry` literal explicitly (none in tree) keeps compiling.

## Scope boundaries

Per the ticket:
- Did **not** touch `src/assets/` (Phase 1 territory — PRs #449/#450)
- Did **not** touch `loadAtlas*` / worker / catalog
- Did **not** implement `scene_assets_acquire/release` hooks (that's #444)
- Did touch the assembler — one extra comptime loop, separate PR linked above

## Test plan

- [x] `zig build test` green
- [x] New engine tests cover: defaults-to-empty for legacy, `registerSceneWithAssets`, `setSceneAssets` happy path + `SceneNotFound`, slash-name round-trip
- [ ] End-to-end wiring verified once labelle-assembler#52 merges

Refs: #445, #437

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -110,6 +110,13 @@ pub fn GameConfig(
         pub const SceneEntry = struct {
             loader_fn: *const fn (*Self) anyerror!void,
             hooks: SceneHooks,
+            /// Declared asset manifest for this scene — the slice of asset names
+            /// the scene needs loaded before it runs. Populated by the assembler
+            /// from each scene's `"assets": [...]` block (see Asset Streaming
+            /// RFC #437 / issue #445). Defaults to an empty slice for scenes
+            /// registered via the legacy (non-manifest) path; scripts can then
+            /// do `game.scenes.get("main").?.assets` without a null check.
+            assets: []const []const u8 = &.{},
         };
 
         /// Runtime JSONC scene path info.
@@ -634,6 +641,8 @@ pub fn GameConfig(
         // ── Scene Management (mixin) ─────────────────────────────
         pub const registerScene = SceneMixin.registerScene;
         pub const registerSceneSimple = SceneMixin.registerSceneSimple;
+        pub const registerSceneWithAssets = SceneMixin.registerSceneWithAssets;
+        pub const setSceneAssets = SceneMixin.setSceneAssets;
         pub const setScene = SceneMixin.setScene;
         pub const setSceneAtomic = SceneMixin.setSceneAtomic;
         pub const queueSceneChange = SceneMixin.queueSceneChange;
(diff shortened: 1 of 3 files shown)


apotema:
@copilot review

Issue #446, opened 2026-04-15
feat(assets): refcount unload via loader.free


apotema:
Part of Phase 3 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Wire `loader.free` on refcount-hits-zero for `ready` assets so unused assets free GPU/audio/font resources automatically.

## Acceptance
- [ ] `release(name)` on a `.ready` entry with refcount hitting 0 calls `loader.free(entry)`, transitions state back to `.registered`, clears the decoded payload
- [ ] No-op on `.registered` entries (nothing to free yet)
- [ ] Re-acquire after free re-enqueues a fresh decode
- [ ] Tests: single acquire/release round trip, shared asset held by two scenes only frees after both release, unload-then-reacquire works
- [ ] Depends on Phase 1 + Phase 2

Issue #443, opened 2026-04-15
refactor(assets): port loadAtlas* to pump-driven sync shim


apotema:
Part of Phase 1 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Collapse the existing `RuntimeAtlas.pending` machinery (#434) into `AssetCatalog`. `loadAtlasFromMemory` and `loadAtlasIfNeeded` become pump-driven sync shims so the current caller surface is unchanged.

## Deadlock fix — MUST call pump() inside the wait loop
Without `pump()`, `isReady` never flips and the thread spins forever. Exact shape from RFC §2:

```zig
pub fn loadAtlasIfNeeded(self: *Game, name: []const u8) !void {
    self.assets.acquire(name);
    while (!self.assets.isReady(name)) {
        if (self.assets.lastError(name)) |err| return err;
        self.assets.pump();           // CRITICAL — without this, deadlock
        std.Thread.yield() catch {};
    }
}
```

## Acceptance
- [ ] `RuntimeAtlas.pending` path removed; registrations go through `AssetCatalog.register`
- [ ] `loadAtlasFromMemory` and `loadAtlasIfNeeded` reimplemented as the shims above
- [ ] All existing atlas callers work unchanged (back-compat smoke tests green)
- [ ] Regression test proving the shim terminates (deadlock guard)

## Expected user-visible outcome
**None.** The shim still blocks the main thread — cold-start freeze is unchanged. Phase 2 delivers the UX win.

Issue #442, opened 2026-04-15
feat(assets): pump() with upload budget + zombie drop path


apotema:
Part of Phase 1 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Implement `AssetCatalog.pump()` draining the worker → main result queue, with a per-frame upload budget and zombie-upload discard.

## Acceptance
- [ ] `pump()` drains up to `UPLOAD_BUDGET_PER_FRAME` (initial: 4) results per call
- [ ] For each result:
  - [ ] `refcount == 0` → `loader.drop`, state back to `.registered` (zombie upload discarded)
  - [ ] `result.err` set → `entry.last_error = err`, state `.failed`
  - [ ] otherwise → `loader.upload(entry, decoded)`; on success `.ready`, on error `.failed`
- [ ] Tests covering all three branches (happy, zombie drop, upload failure)

## Open questions folded in
- [ ] Measure `UPLOAD_BUDGET_PER_FRAME = 4` on Galaxy Tab A7 to confirm no hitch (RFC Open Questions §2)
- [ ] `Game.asset_failure_policy` enum (`fatal` | `warn` | `silent`), default `fatal` — consumed by scene transition; document coupling here (RFC Open Questions §3)

Issue #441, opened 2026-04-15
feat(assets): audio + font stub loaders


apotema:
Part of Phase 1 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Add `src/assets/loaders/audio.zig` and `src/assets/loaders/font.zig` as stub loaders whose `decode()` returns `error.LoaderNotImplemented`.

## Why stubs matter now
Exercising the full register → queue → pump → fail path end-to-end means that when real `decodeAudio` / `decodeFont` land in Phase 4 they only need to provide vtable functions — no streaming machinery work.

## Acceptance
- [ ] `audio.zig` vtable: `decode` returns `error.LoaderNotImplemented`; `upload`/`drop`/`free` are safe no-ops
- [ ] `font.zig` vtable: same shape
- [ ] **NO panics anywhere** — a typo'd manifest entry must not crash the process; it must flow through `.failed` + `lastError`
- [ ] Integration test: register a stub audio asset, `acquire`, pump, assert state is `.failed` and `lastError == error.LoaderNotImplemented`

apotema:
Closing as already satisfied. The audio/font stub vtables returning  landed with the  scaffolding in #449 and have been exercised end-to-end by every subsequent Phase 1 ticket (#450 worker, #452 image loader, #453 pump). Real decoders are tracked separately via #447 (audio) and #448 (font).

PR #452, opened 2026-04-16
feat(assets): image loader — real decodeImage/uploadTexture vtable

### LINKED ISSUE #440: feat(assets): AssetLoaderVTable + image loader
Part of Phase 1 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Add `src/assets/loader.zig` (vtable definition) and `src/assets/loaders/image.zig` (first concrete loader).

## Acceptance
- [ ] `AssetLoaderVTable { decode, upload, drop, free }` per RFC §2:
  - `decode` — worker-thread, returns allocator-owned `DecodedPayload`
  - `upload` — main-thread finalise; frees the CPU payload on success
  - `drop` — main-thread discard path (refcount hit zero between decode and upload); frees CPU payload
  - `free` — unload path (refcount hit zero on `ready`); releases GPU/audio/font resource
- [ ] `DecodedPayload` as a tagged union `{ image, audio, font }` (audio/font empty placeholders for now)
- [ ] `loaders/image.zig` delegates to `gfx.decodeImage` / `gfx.uploadTexture`
- [ ] Tests via mock gfx backend exercising decode → upload → free
- [ ] Depends on labelle-gfx decodeImage/uploadTexture ticket

apotema:
Closes #440. Part of the Asset Streaming RFC (#437) — Phase 1.

## Summary

Replaces the `src/assets/loaders/image.zig` stub (scaffolded in #449) with a real vtable that CPU-decodes on the asset worker thread and GPU-uploads on the main thread, backed by a runtime-injected `ImageBackend` hook that the assembler populates at `Game.init` with adapters for `labelle-gfx`'s `decodeImage` / `uploadTexture` / `unloadTexture` (merged in labelle-gfx#245 and labelle-assembler#51).

After this lands, the ground is prepared for #442 (pump() body), #443 (legacy atlas shim), and #444–#446 (scene-transition acquire/release) — every one of them can now assume a loader that actually produces ready GPU handles.

## Backend wiring — why a runtime hook, not a direct gfx import

The engine crate does **not** depend on `labelle-gfx`. `build.zig` only wires `labelle-core`, `scene` and `jsonc` into the engine module, and that is deliberate: `labelle-gfx/build.zig.zon` still carries a path-relative dep on `../labelle-core` which breaks URL fetches (see the #440 ticket prologue), and we do not want the engine to inherit that fragility.

So instead of `@import(\"labelle-gfx\")`, `loaders/image.zig` exposes a tiny `ImageBackend` struct of function pointers:

\`\`\`zig
pub const ImageBackend = struct {
    decode: *const fn ([:0]const u8, []const u8, Allocator) anyerror!DecodedImage,
    upload: *const fn (DecodedImage) anyerror!Texture,
    unload: *const fn (Texture) void,
};
\`\`\`

The assembler (which already imports labelle-gfx) calls `engine.ImageLoader.setBackend(...)` at `Game.init` with adapters that forward to `gfx.decodeImage` / `gfx.uploadTexture` / `gfx.unloadTexture`. Until that call lands, `decode`/`upload` return `error.ImageBackendNotInitialized` — a clean error rather than a null deref. The assembler-side wiring is a separate follow-up; this PR ships only the engine side.

This approach also **side-steps the `DecodedImage` nominal-type collision** described in the ticket prologue entirely: because the engine never imports labelle-gfx's `DecodedImage`, there is no way for two structurally-identical types to collide through the generic `Backend(Impl)` wrapper. The backend adapter on the assembler side is the only place that has to know about both.

## Where does the uploaded Texture live? — option 1

Picks option 1 from the ticket — extends `AssetEntry` with a parallel \`resource: ?UploadedResource\` field alongside the existing \`decoded: ?DecodedPayload\`:

\`\`\`zig
pub const UploadedResource = union(LoaderKind) {
    image: Texture,   // u32 — matches src/atlas.zig's convention
    audio: struct {}, // placeholder for #441
    font:  struct {}, // placeholder for #441
};
\`\`\`

This mirrors the RFC §2 sketch (\`decoded: union { texture, audio, font }\`), survives #441 / #442 / #446 cleanly, and keeps the CPU-side and GPU-side slots independent. Option 2 (reusing \`decoded\` as a before-OR-after union) was rejected because Zig unions are comptime-tagged; option 3 (private loader state) adds another indirection nobody else needs.

The decision is documented in a top-of-file comment on \`src/assets/loader.zig\`.

## Ownership of \`DecodedImage.pixels\`

- \`decode\` allocates the pixel buffer through the worker's allocator.
- \`upload\` hands the pixels to the GPU and then \`allocator.free\`s them (the labelle-gfx contract explicitly says \`uploadTexture\` does NOT take ownership). On the error path \`upload\` leaves the buffer alive so the caller can route it to \`drop\`.
- \`drop\` (refcount-zero during decode) frees the CPU buffer, never touches the GPU.
- \`free\` (refcount-zero on \`.ready\`) releases the texture via \`backend.unload\` and clears \`entry.resource\`.

The vtable's \`upload\` signature gains an \`allocator: Allocator\` parameter so it can free on the success path; the audio / font stubs take the new parameter and \`_ = allocator\` it.

## Tests

Six in-source unit tests on \`loaders/image.zig\` via a recording mock backend:

- no-backend error
- full round-trip (decode → upload → free)
- drop path (refcount hit zero mid-decode)
- upload error leaves pixels alive for \`drop\` to clean up
- no-op \`free\` without a backend or resource
- decode error propagation

Two integration tests in \`test/asset_catalog_test.zig\` via a second mock:

- end-to-end \`register → acquire → worker-decode → manual upload → release → free\` round-trip
- catalog discard path — refcount hits zero between decode and upload, and \`deinit\` drains the orphaned payload through \`vtable.drop\`

Both run under \`testing.allocator\` (a GPA in test mode), which flags leaks **and** double-frees.

Updates the previously-passing \`catalog.zig\` + \`worker.zig\` tests to expect \`error.ImageBackendNotInitialized\` (new) instead of \`error.NotImplemented\` (stub) and to \`clearBackend()\` up front so a prior test cannot leak a mock.

## Build plumbing

\`zig build test\` gained a new test step rooted at \`src/assets/mod.zig\`. The existing \`test/asset_catalog_test.zig\` binary lives in its own module and reaches the engine through a cross-module import — Zig only discovers in-source \`test\` blocks in files that belong to the SAME module as the test binary's root, so every \`test\` under \`src/assets/\` (including the catalog / worker tests from #449 / #450 that were silently not running in CI) needs a dedicated root to execute.

**Net test count: 125 → 151.** That is +6 image loader unit tests, +2 image loader integration tests, and +18 latent catalog/worker tests from prior PRs that are now actually running.

## Verification

- \`zig build test\` — 151/151 passing
- \`zig test src/assets/mod.zig\` — 24/24 passing (catalog 12 + worker 5 + image 6 + test root 1)
- Direct compile of \`test/asset_catalog_test.zig\` — 5/5 passing

## Test plan

- [x] \`zig build test\` green on macOS arm64
- [ ] CI green on the PR
- [ ] Follow-up ticket (assembler / codegen template) wires \`gfx.decodeImage\` / \`gfx.uploadTexture\` / \`gfx.unloadTexture\` through \`engine.ImageLoader.setBackend\` at \`Game.init\`
- [ ] #442 picks up the new vtable and drives it from \`pump()\`

Refs: #437 (RFC), #449 (catalog scaffold), #450 (worker + rings)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -60,4 +60,22 @@ pub fn build(b: *std.Build) void {
         });
         test_step.dependOn(&b.addRunArtifact(t).step);
     }
+
+    // In-source tests under `src/assets/` (catalog, worker, loader,
+    // loaders/*) cannot be dragged into a `test/*` binary through the
+    // cross-module `engine` import — Zig only discovers top-level
+    // `test` blocks in files that belong to the **same** module as
+    // the test binary's root. Rooting an extra test binary directly
+    // at `src/assets/mod.zig` gives those files a module of their
+    // own where `_ = @import("...")` chains actually cascade the
+    // test blocks into the binary. Added as part of #440 while the
+    // real image loader started writing its own in-source tests.
+    const assets_tests = b.addTest(.{
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("src/assets/mod.zig"),
+            .target = target,
+            .optimize = optimize,
+        }),
+    });
+    test_step.dependOn(&b.addRunArtifact(assets_tests).step);
 }
--- a/src/assets/catalog.zig
+++ b/src/assets/catalog.zig
@@ -33,6 +33,8 @@ const worker_mod = @import("worker.zig");
 
 pub const LoaderKind = loader_mod.LoaderKind;
 pub const DecodedPayload = loader_mod.DecodedPayload;
+pub const UploadedResource = loader_mod.UploadedResource;
+pub const Texture = loader_mod.Texture;
 pub const AssetLoaderVTable = loader_mod.AssetLoaderVTable;
 pub const AssetState = loader_mod.AssetState;
 pub const AssetEntry = loader_mod.AssetEntry;
@@ -138,6 +140,7 @@ pub const AssetCatalog = struct {
             .raw_bytes = bytes,
             .file_type = file_type,
             .decoded = null,
+            .resource = null,
             .last_error = null,
         });
     }
@@ -289,7 +292,8 @@ test "register then acquire bumps refcount and enqueues work" {
     try testing.expectEqual(@as(u32, 1), entry.refcount);
     // First acquire moved the entry to `.queued` — the worker ring
     // has taken ownership of the request and will eventually publish
-    // an `error.NotImplemented` result on the stub loader.
+    // an `error.ImageBackendNotInitialized` result on the real image
+    // loader (no backend injected in the unit-test harness).
     try testing.expectEqual(AssetState.queued, entry.state);
     try testing.expectEqual(LoaderKind.image, entry.loader_kind);
     try testing.expectEqual(@as(?DecodedPayload, null), entry.decoded);
@@ -408,7 +412,13 @@ test "pump is a no-op until the worker lands (#442)" {
     try testing.expect(!catalog.isReady("background"));
 }
 
-test "acquire spawns worker which produces NotImplemented for stub loader" {
+test "acquire spawns worker which surfaces ImageBackendNotInitialized without a backend" {
+    // Make sure no previous test left a backend injected on this
+    // process-global slot — the assertions below rely on the loader
+    // returning the not-initialised error, not a mock success.
+    const image_loader_mod = @import("loaders/image.zig");
+    image_loader_mod.clearBackend();
+
     var catalog = AssetCatalog.init(testing.allocator);
     defer catalog.deinit();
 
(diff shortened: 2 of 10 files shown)


apotema:
@copilot review

PR #457, opened 2026-04-16
refactor(assets): port loadAtlas* to pump-driven sync shim

### LINKED ISSUE #443: refactor(assets): port loadAtlas* to pump-driven sync shim

apotema:
## Summary

Routes the legacy `Game.registerAtlasFromMemory` / `loadAtlasFromMemory` / `loadAtlasIfNeeded` / `isAtlasLoaded` family through the `AssetCatalog` + asset worker thread instead of calling `renderer.loadTextureFromMemory` directly on the main thread. The surface is unchanged — every existing caller (notably the smoke-test example's assembler-generated `main.zig`) keeps compiling and running — but the PNG decode now happens off-thread.

Adds `Game.assets: AssetCatalog` (initialised alongside `atlas_manager` in `init`, torn down after it in `deinit` so the worker joins while the allocator is still alive).

### The load-bearing pump

Per RFC §2, `loadAtlasIfNeeded` is now a pump-driven sync shim:

```zig
_ = try self.assets.acquire(name);
while (!self.assets.isReady(name)) {
    if (self.assets.lastError(name)) |err| return err;
    self.assets.pump();           // load-bearing
    std.Thread.yield() catch {};
}
```

Without the `pump()` call the main thread never finalises the upload and the loop spins forever. The deadlock regression test traps that failure inside a 200ms deadline on a background thread so CI fails loudly instead of stalling.

### Scope

- `registerAtlasFromMemory` keeps the legacy `TextureManager.registerPendingAtlas` side-effects (JSON parsed eagerly, `meta.size` cached) and additionally mirrors the PNG onto the catalog. Double-register (when assembler-generated init code already registered the same name from a scene manifest) swallows `error.AssetAlreadyRegistered`.
- `loadAtlasFromMemory` becomes `registerAtlasFromMemory` + `loadAtlasIfNeeded`. **Still blocks the calling frame — deliberate no cold-start UX change.** Spreading the decode across frames is Phase 2's job.
- `isAtlasLoaded` still reads from `TextureManager`, not the catalog: the atlas is "loaded" from the engine's point of view only after `markPendingLoaded` seeds `RuntimeAtlas.texture_id`.
- `RuntimeAtlas.pending` (engine #434) is partially retired — `pending.bytes` / `pending.file_type` are no longer consumed, but `pending.meta` is still used by `applyTextureScale`. Full retirement can follow once Phase 2 lands.

### Known behavioural change

Atlases whose JSON `meta.size` differs from the actual PNG pixel dims no longer get an automatic `texture_scale`. The catalog-managed upload path bypasses `renderer.loadTextureFromMemory`, so the renderer's texture side-table is empty and `getTextureInfo` returns null. Only affects projects that shipped a downscaled PNG without re-running TexturePacker — the common 1:1 case is unchanged. An explicit workflow lands with Phase 2.

### Closes & refs

Closes #443.
Related to #437 (RFC).

## Test plan

- [x] `zig build test` — 168/168 tests pass (161 pre-existing + 7 new in `test/asset_streaming_shim_test.zig`):
  - [x] `registerAtlasFromMemory` + `loadAtlasIfNeeded` round-trip
  - [x] `loadAtlasFromMemory` eager surface contract (atlas ready on return)
  - [x] double-call idempotence (second `loadAtlasIfNeeded` is a no-op)
  - [x] `AtlasNotFound` on unknown names
  - [x] `isAtlasLoaded` pre/post load
  - [x] deadlock regression — `error.MockDecodeFailure` surfaces via `lastError` within 200ms, no hang
  - [x] double-register tolerated (manifest path + legacy shim coexist)
- [ ] Generate the smoke-test example via labelle-assembler and verify the emitted `main.zig` still compiles + runs against this engine branch (manual step — cross-repo).

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -42,6 +42,7 @@ pub fn build(b: *std.Build) void {
         "test/jsonc_bridge_leak_test.zig",
         "test/scene_ref_test.zig",
         "test/asset_catalog_test.zig",
+        "test/asset_streaming_shim_test.zig",
         "test/animation_def_test.zig",
     };
 
--- a/src/assets/catalog.zig
+++ b/src/assets/catalog.zig
@@ -290,6 +290,25 @@ pub const AssetCatalog = struct {
         return entry.last_error;
     }
 
+    /// Rewind a `.failed` entry back to `.registered` so the next
+    /// `acquire` re-enqueues the decode. Without this, a transient
+    /// decode/upload failure becomes permanent: `acquire` only
+    /// enqueues from `.registered`, so a retry would re-bump the
+    /// refcount and immediately surface the stale `last_error` from
+    /// the failed attempt without re-triggering work.
+    ///
+    /// Caller-driven (not automatic) so the failure is observable
+    /// via `lastError` first — letting the caller decide whether to
+    /// retry or surface the error to the user. No-op for entries
+    /// that aren't `.failed`. Clears `last_error` so a subsequent
+    /// `lastError` call returns null.
+    pub fn resetFailed(self: *AssetCatalog, name: []const u8) void {
+        const entry = self.entries.getPtr(name) orelse return;
+        if (entry.state != .failed) return;
+        entry.state = .registered;
+        entry.last_error = null;
+    }
+
     /// Drains worker results and finalises uploads on the main thread.
     /// Caps itself at `UPLOAD_BUDGET_PER_FRAME` so a burst of ready
     /// decodes cannot stall a frame; the next `pump()` picks up where
--- a/src/game.zig
+++ b/src/game.zig
@@ -1158,6 +1158,15 @@ pub fn GameConfig(
             // entry enqueues the decode; subsequent acquires just pin
             // the refcount so the zombie-drop path in `pump()` can't
             // rewind us while we are waiting for the upload to land.
+            //
+            // `errdefer release` guarantees the shim returns the
+            // refcount on every failure path (lastError, missing
+            // entry, wrong asset kind, markPendingLoaded error, …).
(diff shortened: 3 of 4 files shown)


apotema:
@copilot review

PR #463, opened 2026-04-17
docs: RFC for audio loader — Phase 4 of asset streaming (#447)

### LINKED ISSUE #447: tracking: audio loader — Phase 4 of asset streaming
Tracking issue for Phase 4 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Real `decodeAudio` + device handle management replaces the stub loader landed in Phase 1. Separate RFC required per the RFC Migration §Phase 4.

## Out of scope for this ticket
Streaming machinery is already done by the time Phase 4 starts. This ticket only needs:
- Audio backend decoder (stb_vorbis / dr_wav / format TBD)
- Audio device handle lifecycle on `upload` / `free`
- Manifest support for audio names (already validated by the assembler asset-name validation ticket)

## Next step
Draft follow-up RFC before implementation.

apotema:
Draft design document for replacing the stub audio loader with a real decoder + backend hook. Seeking feedback before implementation starts.

## Summary
- Mirrors the image-loader pattern from Phase 2 — `AudioBackend` runtime hook struct, injected by the assembler at `Game.init`, same threading model as `ImageBackend`.
- Single-header decoders: **dr_wav** for WAV, **stb_vorbis** for OGG. MP3 deferred. Linked into the backend, not the engine.
- Signed 16-bit interleaved PCM at source sample rate / channel count. No resampling at decode time.
- Fleshes out the placeholder `DecodedPayload.audio` / `UploadedResource.audio` variants.
- Extension-driven codegen at the assembler side.
- Failure path reuses `asset_failure_policy` from #444 — a corrupt file cannot crash.

## Migration
Three landable commits, each green:
1. Engine: types + vtable + mock-tested unit test.
2. Assembler: file-type dispatch for `.wav` / `.ogg`.
3. Backend: one PR per audio backend (raylib-audio, sokol-audio, …).

## Open questions (in the doc)
1. Canonical sample format at decode vs preserve source.
2. Byte-size memory budget on `AssetEntry` — now or later.
3. Streamed music: share the catalog (new `.music` LoaderKind) or keep it on the existing Music interface.

Leaning toward the answers proposed in the doc, but each is genuinely uncertain — happy to flip.

Parent: [RFC-ASSET-STREAMING §Phase 4](../blob/main/RFC-ASSET-STREAMING.md#migration-plan).
Closes #447 on acceptance; the tracking issue stays open until all three migration PRs land.

### DIFF
--- a/RFC-AUDIO-LOADER.md
+++ b/RFC-AUDIO-LOADER.md
@@ -69,8 +69,10 @@ pub const DecodedAudio = struct {
 
 Mirrors `ImageBackend` exactly. Engine never imports the audio crate; the assembler injects adapters at `Game.init`.
 
+The handle the backend returns is the existing `audio_types.SoundId` (`{ index: u32, generation: u32 }`, already used across `src/audio.zig` with `isValid` + generation-based staleness detection). Defining a second `u32` SoundId here would collide with that type and split the engine's sound-handle vocabulary in two — reviewers flagged this as an API-consistency hazard on the first draft. Keep the existing struct; the backend can build one from a raw backend integer via `.{ .index = raw, .generation = … }`.
+
 ```zig
-pub const SoundId = u32;  // opaque backend handle, parallels Texture
+const audio_types = @import("audio_types.zig");  // existing engine module
 
 pub const AudioBackend = struct {
     decode: *const fn (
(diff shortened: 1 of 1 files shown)


apotema:
@copilot review — addressed all six comments. They converged on three substantive concerns:

1. **SoundId conflict (gemini L73 + copilot L75)**: dropped the new `u32 SoundId` and reused the existing `audio_types.SoundId` struct (`{ index, generation }`) so the engine doesn't end up with two handle types for the same concept.

2. **Circular dep in loader.zig variants (copilot L112/L114)**: `audio: audio_loader.DecodedAudio` / `audio: audio_loader.SoundId` would point `loader.zig` at its own child module. Adopted the image-loader pattern: inline PCM struct in `DecodedPayload`, `audio_types.SoundId` in `UploadedResource`.

3. **`free` early-return broke cleanup signal (copilot L144)**: matched `image.zig`'s pattern — unconditionally clear `entry.resource = null`, only gate the `backend.unload` call.

Bonus: gemini L130 `upload` signature concern was a false positive (upstream RFC-ASSET-STREAMING is stale; merged `AssetLoaderVTable.upload` in `src/assets/loader.zig` does take `allocator`). Added a closing note pointing readers at the authoritative vtable definition.

apotema:
@copilot review

PR #464, opened 2026-04-17
fix(assets): compile under single_threaded (WASM/emscripten) (#461)

### LINKED ISSUE #461: Asset pipeline fails to compile on single-threaded targets (WASM)
## Problem

`AssetCatalog.ensureWorker` unconditionally calls `std.Thread.spawn` (via `worker_mod.AssetWorker.start()` → `worker.zig:181`). `std.Thread.spawn` is a **hard compile error** under `single_threaded = true` (Zig stdlib `Thread.zig:439`), so any build that enables single-threaded mode — notably WASM builds — fails to compile the moment the asset catalog's type is reached.

## Symptom

```
/.../std/Thread.zig:439:9: error: Cannot spawn thread when building in single-threaded mode
    start: .../labelle-engine/src/assets/worker.zig:181:43
    ensureWorker: .../labelle-engine/src/assets/catalog.zig:93:30
```

## Impact

- **labelle-cli** `Docker Build Test` CI job has been red on `main` for some time ([example run on `main`](https://github.com/labelle-toolkit/labelle-cli/actions/runs/24523334563)).
- Any downstream project that targets WASM with single-threaded mode will hit the same error.

## History

- Present before the [parallel-worker refactor (PR #460)](https://github.com/labelle-toolkit/labelle-engine/pull/460) — that PR did not introduce the failure, it only increased the number of spawn call sites (from 1 to `NUM_WORKERS`).
- The `worker.zig`/`catalog.zig` module pair has always assumed a threaded host.

## Fix options

Pick one — each has a different impact on WASM runtime behavior:

1. **Synchronous fallback.** When `builtin.single_threaded`, skip `ensureWorker` and decode synchronously inside `acquire`. Simpler API-wise: every asset blocks the main thread on first use. Cold start regresses on WASM but the pipeline works.
2. **Stubbed-out worker.** `ensureWorker` becomes a no-op on single-threaded, `acquire` still enqueues, and a new `drainSynchronously` helper (called from `pump`?) dequeues + decodes + uploads in-line. Cleaner separation between "enqueue" and "work"; still single-threaded work on main.
3. **Compile-time disable.** Make the catalog's public surface error out on single-threaded (`@compileError`) and require callers to avoid it. Only tenable if WASM apps never touch assets, which is unlikely.

Option (1) or (2) is probably the right path. Needs a design thought pass — particularly around `upload` which already runs on the main/GL thread.

## Related

- labelle-cli [PR #175](https://github.com/labelle-toolkit/labelle-cli/pull/175) surfaced this during review (Docker Build Test red).
- labelle-engine [PR #460](https://github.com/labelle-toolkit/labelle-engine/pull/460) is the parallel-decode work that sat on top of this existing limitation.

apotema:
Closes #461. Unblocks the \`Docker Build Test\` CI job on labelle-cli, which has been red on main since the asset worker landed.

## Summary
- \`AssetWorker.start\` no-ops under \`builtin.single_threaded\` instead of calling the unconditional \`std.Thread.spawn\` (the hard compile error the ticket describes).
- New \`AssetWorker.runOnce\` drains every pending request on the main thread via the same decode + publish path the worker thread takes — \`decodeAndPublish\` and \`buildResult\` helpers extracted so there's no behavioural divergence between threaded and sync drains.
- \`AssetCatalog.acquire\` calls \`workers[idx].runOnce()\` after enqueueing under \`single_threaded\`, so the result lands in the result ring before the next \`pump()\`. Threaded builds are bit-for-bit unchanged.

## Which fix option (from the ticket)
**Option 1** — synchronous fallback. Cold start regresses on WASM because each asset blocks the main thread on first use, but the pipeline works end-to-end. Option 2 (explicit \`drainSynchronously\` decoupled from \`acquire\`) is a nicer design but is pure polish on top of a working baseline; easier to land incrementally.

## Tests
- **New compile-only build step** — \`assets_single_threaded\` in \`build.zig\`, targets \`src/assets/mod.zig\` with \`single_threaded = true\`. Catches any future unguarded \`Thread.spawn\` at \`zig build\` time with no runtime needed; the original regression was a type-checker failure.
- **New in-source test**: \`AssetWorker.runOnce drains a request synchronously\`. Never calls \`start()\` (matches WASM) and asserts the request flows through decode + publish.
- Existing \`SpscRing\` producer-thread test skipped under \`single_threaded\` with \`error.SkipZigTest\`.

## Test plan
- [x] \`zig build\` clean.
- [x] \`zig build test\` — 186/186 on macOS arm64, 42/42 build steps including the new single-threaded compile guard.
- [ ] labelle-cli \`Docker Build Test\` green after this lands + flying-platform bumps to the next engine release.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -86,8 +86,9 @@ pub fn build(b: *std.Build) void {
     // Historically `std.Thread.spawn` was unconditional and this
     // combination was a hard compile error on `main`, breaking every
     // downstream WASM build. A compile-only check here catches any
-    // regression at `zig build` time — no runtime needed because the
-    // point is the type-checker reaching `std.Thread.spawn`.
+    // regression at `zig build test` time (the step it's wired to) —
+    // no runtime needed because the point is the type-checker
+    // reaching `std.Thread.spawn`.
     const assets_single_threaded = b.addTest(.{
         .root_module = b.createModule(.{
             .root_source_file = b.path("src/assets/mod.zig"),
--- a/build.zig
+++ b/build.zig
@@ -80,4 +80,21 @@ pub fn build(b: *std.Build) void {
         }),
     });
     test_step.dependOn(&b.addRunArtifact(assets_tests).step);
+
+    // Issue #461 regression guard: the asset pipeline must compile
+    // under `single_threaded = true` (WASM / emscripten default).
+    // Historically `std.Thread.spawn` was unconditional and this
+    // combination was a hard compile error on `main`, breaking every
+    // downstream WASM build. A compile-only check here catches any
+    // regression at `zig build` time — no runtime needed because the
+    // point is the type-checker reaching `std.Thread.spawn`.
+    const assets_single_threaded = b.addTest(.{
+        .root_module = b.createModule(.{
+            .root_source_file = b.path("src/assets/mod.zig"),
+            .target = target,
+            .optimize = optimize,
+            .single_threaded = true,
+        }),
+    });
+    test_step.dependOn(&assets_single_threaded.step);
 }
--- a/src/assets/catalog.zig
+++ b/src/assets/catalog.zig
@@ -26,6 +26,7 @@
 //!    `project.labelle`.
 
 const std = @import("std");
+const builtin = @import("builtin");
 const Allocator = std.mem.Allocator;
 
 const loader_mod = @import("loader.zig");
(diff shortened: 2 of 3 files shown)


apotema:
@copilot review

apotema:
Pushed `11282a2` on top of copilot's `6e905bc`. Addresses the remaining review comments:

**Gemini / Copilot HIGH — result-ring-full stranding.** `runOnce` now calls `results.isFull()` BEFORE dequeueing a request; a full result ring leaves the work in the request ring for a later `runOnce` to pick up. New `SpscRing.isFull()` helper mirrors `isEmpty`.

**Gemini medium — misleading `.sync` branch comment.** Rewritten to describe the branch as a defensive fallback that shouldn't fire given the new `isFull` pre-check; log level bumped from debug to warn.

**Copilot medium — `start()` assertion order.** Moved the `self.thread == null` assert in front of the `single_threaded` early return, so the "exactly once" contract is enforced consistently across targets. (Copilot's commit added a parallel assert on `runOnce`.)

**Copilot medium — stub-request test hangs on single-threaded.** Added `if (builtin.single_threaded) return error.SkipZigTest;` — the threaded path stays covered here and the single-threaded path has its own `runOnce` test below.

**Copilot medium — build.zig comment wording.** `zig build` → `zig build test` (matches the step the guard is wired to).

Combined with copilot's `6e905bc` (comptime mode, runOnce assert): 42/42 build steps, 186/186 tests pass on macOS arm64.

PR #466, opened 2026-04-18
feat(engine): plugin-accessible pause flag + pause_changed hook

### LINKED ISSUE #465: engine: plugin-accessible pause hook — let Controller.advance gate on pause without knowing about game components
## Problem

Plugin Controllers (RFC https://github.com/Flying-Platform/flying-platform-labelle/blob/main/docs/RFC-plugin-controllers.md) expose per-frame methods (pathfinder's \`advance\`, asset-catalog's \`pump\`, etc.) that plugin-shipped scripts invoke. Today there's no way for those plugin scripts to check whether the game is paused — the concept is game-side (a \`GamePaused\` component in flying-platform-labelle) and the plugin cannot import from the game's tree (Zig module-path boundary correctly rejects it).

Concrete symptom: the pathfinder migration PR (Flying-Platform/flying-platform-labelle#216) had to yank its plugin-shipped \`01_advance.zig\` and move the \`Controller.advance\` call into the game-side dispatch script, where the pause component is locally importable. This works as a stopgap but walks back one of the architectural wins of the plugin-controllers pattern — the per-frame entry point was supposed to live with the plugin.

Every future plugin with per-frame work will hit this the same way (asset catalog pump, scheduler flush queue, behavior tree step, etc.).

## Proposal

The engine exposes a plugin-accessible pause primitive. Shape options:

### Option A — \`game.isPaused() bool\`

Simplest. Engine holds a single \`paused: bool\` flag on \`Game\`. A game opts in by adding a \`GamePaused\` component (or equivalent) and syncing its presence to the engine flag in one top-level script (\`scripts/*.zig\`, runs in all states) — or the engine could adopt \`GamePaused\` as a convention and sync internally.

Plugin scripts call \`if (game.isPaused()) return;\` — no component import needed, no cross-repo type coupling.

### Option B — pause hook registration

Plugins register a callback the engine invokes with the current pause state on each tick. More flexible but heavier (hook plumbing), and the plugin-scripts approach (where the plugin just ships a \`.zig\` script the engine ticks) doesn't need this indirection.

### Option C — \`game_states\` extension

Extend the \`game_states\` convention so a plugin can opt out of ticking during arbitrary states. Game-wise, \`paused\` becomes a pseudo-state. Clean model but requires every game that pauses to pick a \`paused\` state name, and existing games using a component-based pause have to migrate.

### Leaning

Option A — smallest surface, matches the existing \`game\` parameter contract for plugin scripts, no convention-name negotiation. \`GamePaused\` can stay as a game-side component; the game just adds a tiny \`scripts/pause_sync.zig\` (or the engine learns about \`GamePaused\` by convention name) to keep the engine flag in sync.

## Scope

- [ ] Add \`Game.paused: bool\` + \`isPaused()\` / \`setPaused(bool)\` methods in labelle-engine's \`src/game.zig\`
- [ ] Document in the plugin-controllers RFC that plugin-shipped scripts should gate on \`game.isPaused()\` instead of reaching into game components
- [ ] Update flying-platform-labelle to sync \`GamePaused\` → \`Game.paused\` in a tiny always-on script (or propose the \`GamePaused\` naming convention for automatic engine handling)
- [ ] Restore the plugin-shipped \`libs/pathfinder/scripts/playing/01_advance.zig\` that PR #216 had to delete; move \`Controller.advance\` back out of the game-side dispatch

## Cross-repo

Engine change here; consumer updates in flying-platform-labelle (once a new labelle-engine release picks up the change).

## Priority

Not blocking any specific feature. Every new plugin-controllers migration that wants per-frame work either:
- ships an advance script that ignores pause (buggy under the game's pause UX), or
- has its advance call moved into the game-side dispatch (violates the \"plugin owns its per-frame work\" architectural win)

Worth landing before the second plugin-controllers migration (worker_controller / scheduler / caretaker) hits the same wall.

## Related

- https://github.com/Flying-Platform/flying-platform-labelle/pull/216 — pathfinder migration that surfaced this, cursor bugbot review round 1
- https://github.com/Flying-Platform/flying-platform-labelle/blob/main/docs/RFC-plugin-controllers.md — RFC §2 "plugin-shipped per-frame scripts"
- Flying-Platform/flying-platform-labelle#208 — Plugin Controllers tracking

apotema:
## Summary

- Adds `Game.paused: bool` (default `false`), `setPaused(bool)`, and `isPaused()` so plugin-shipped scripts (pathfinder `Controller.advance`, asset catalog `pump`, etc.) can gate per-frame work on pause without importing a game-side `GamePaused` component across Zig module boundaries.
- Adds a `pause_changed: { paused: bool }` variant to the engine `HookPayload` union; `setPaused` emits it only on transition and is idempotent when the value is unchanged.
- Wires the legacy `pause()` / `resume_()` helpers through `setPaused` so the existing `time_scale`-driven path stays consistent with the new flag (and the duplicate `time_scale == 0` flavour of `isPaused` goes away).

This is item 1 of the four-item scope in the issue. Items 2-4 (RFC docs update, game-side `GamePaused` -> engine sync script, restoring the plugin-shipped `01_advance.zig`) live in the consumer / docs repos and follow up after a release that picks this commit up.

## Test plan

- [x] `zig build test` in `labelle-engine` (184/184 tests pass; 6 new tests in `test/pause_hook_test.zig`)
- [ ] Once consumed: pathfinder plugin's restored `Controller.advance` script gates correctly on `game.isPaused()` in flying-platform-labelle

Closes labelle-toolkit/labelle-engine#465

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -45,6 +45,7 @@ pub fn build(b: *std.Build) void {
         "test/asset_streaming_shim_test.zig",
         "test/animation_def_test.zig",
         "test/scene_assets_hooks_test.zig",
+        "test/pause_hook_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/game.zig
+++ b/src/game.zig
@@ -259,6 +259,15 @@ pub fn GameConfig(
         /// Time scale factor: 0 = paused, 0.5 = slow-mo, 1.0 = normal, 2.0 = fast.
         /// When paused (0), rendering and GUI continue but tick logic stops.
         time_scale: f32 = 1.0,
+        /// Plugin-accessible pause flag (#465). Plugin-shipped scripts
+        /// (pathfinder's `Controller.advance`, asset-catalog's `pump`,
+        /// etc.) gate on `game.isPaused()` instead of importing a
+        /// game-side `GamePaused` component, which would cross the Zig
+        /// module-path boundary. The game owns the source of truth —
+        /// typically a tiny always-on script syncing its own
+        /// `GamePaused` component into `setPaused` — and the engine
+        /// just stores + dispatches the bool.
+        paused: bool = false,
 
         // Profiling (debug builds only)
         /// Opaque pointer to ScriptRunner's profile array. Set by generated code.
@@ -748,6 +757,34 @@ pub fn GameConfig(
         pub const queueStateChange = StateMixin.queueStateChange;
         pub const getState = StateMixin.getState;
 
+        // ── Pause flag (#465) ───────────────────────────────────────
+
+        /// Set the engine-level pause flag. Emits `pause_changed` when
+        /// the value transitions; idempotent if already at `paused`.
+        ///
+        /// Intended to be called from a game-side sync script that
+        /// mirrors a game-owned `GamePaused` component (or equivalent)
+        /// into the engine, so plugin-shipped scripts can gate via
+        /// `game.isPaused()` without importing game components.
+        pub fn setPaused(self: *Self, paused: bool) void {
+            if (self.paused == paused) return;
+            self.paused = paused;
+            self.emitHook(.{ .pause_changed = .{ .paused = paused } });
+        }
+
+        /// Read the engine-level pause predicate. Returns true if
+        /// *either* the explicit `paused` flag is set OR `time_scale`
+        /// is zero — both are valid pause mechanisms and tick gating
+        /// should honour both uniformly.
+        ///
+        /// Plugin scripts should use this instead of reaching for a
+        /// game-side `GamePaused` component — the latter would couple
+        /// plugin code to the game's component types across module
+        /// boundaries.
+        pub fn isPaused(self: *const Self) bool {
+            return self.paused or self.time_scale == 0;
+        }
+
         pub fn unloadCurrentScene(self: *Self) void {
             if (has_events) self.event_buffer.clearRetainingCapacity();
             if (self.current_scene_name) |name| {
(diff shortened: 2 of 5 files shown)


apotema:
Rebased onto \`main\` (resolved one \`build.zig\` test-list conflict with the scene-assets-hooks test that landed in parallel), then folded review fixes into the single feature commit (now \`02d7747\`):

### Addressed

- **gemini HIGH** \`src/game.zig:764\` — pause vs time_scale desync. Resolved by making \`isPaused()\` the unified predicate: returns \`self.paused or self.time_scale == 0\`. Tick now gates on \`self.isPaused()\` instead of \`scaled_dt == 0\`, so an explicit \`setPaused(true)\` halts the tick regardless of time_scale, and a \`setTimeScale(0)\` still pauses through the same predicate. \`pause()\` / \`resume_()\` already sync both flags; \`setPaused\` / \`setTimeScale\` can be called independently and the predicate stays correct.
- **gemini MED** \`test/pause_hook_test.zig:21\` — \`catch {}\` → \`catch unreachable\` in the \`PauseRecorder.pause_changed\` handler. Loud fail on allocation error instead of silent drop.
- **cursor LOW** \`src/hooks_types.zig:67\` — re-exported \`PauseChangedInfo\` from \`src/root.zig\` alongside every other \`*Info\` / \`*Payload\` hook type.

### Hook semantics

\`pause_changed\` still only fires on explicit \`setPaused\` transitions — not when \`setTimeScale\` dips through 0. The hook stays a signal for intentional pause/resume; \`time_scale\` remains a continuous speed knob. If a caller wants hooks on time-scale changes, they can call \`setPaused\` alongside \`setTimeScale(0)\`.

184/184 tests pass.

PR #469, opened 2026-04-22
feat(game): emitSync for synchronous event dispatch

apotema:
## Summary
- Adds `Game.emitSync(event)` — same shape as `emit`, but dispatches the buffered Payload immediately to registered hooks instead of queuing it for end-of-frame `dispatchEvents`.
- Uses the exact same `@unionInit(Payload, @tagName(tag), data)` conversion as `dispatchEvents`, so hook signatures and tags stay identical.

## Why
Motivated by a cross-plugin race in [flying-platform-labelle #280](https://github.com/Flying-Platform/flying-platform-labelle/issues/280): `needs_machine` emits `need_fulfillment_requested` during its controller tick, but the buffered dispatch leaves a window where a sibling plugin (`production`) running later in the same tick sees the worker as idle with no pending fulfillment and double-assigns a delivery. Switching the hunger-dispatch site from `emit` → `emitSync` closes that window without changing the handler signature or breaking the game-script hook contract.

The game-side fix for #280 uses a reservation flag to mitigate the race without this method, but `emitSync` makes the underlying fix available to any plugin that needs synchronous event semantics.

## Test plan
- [x] `zig build test` — existing suite passes
- [x] `emitSync` is `pub fn`; no change to `emit`/`dispatchEvents` semantics
- [ ] Once this ships, flying-platform-labelle will switch the hunger dispatch to `emitSync` and the reservation can be narrowed

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -398,8 +398,32 @@ pub fn GameConfig(
         /// caller needs the handler to have run before the next
         /// statement (cross-plugin state machines that can't tolerate
         /// the buffered-dispatch window).
+        ///
+        /// ## Caveats
+        ///
+        /// The event-buffer design in the custom-game-events RFC exists
+        /// precisely to avoid these, so reach for `emitSync` only when
+        /// the buffered path is provably wrong for the call site:
+        ///
+        /// - **Re-entrancy.** Handlers run mid-tick, inside whatever
+        ///   script or plugin called `emitSync`. A handler that itself
+        ///   mutates entity state, emits more events, or calls back
+        ///   into the caller's own code can interleave with partially-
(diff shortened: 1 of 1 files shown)


apotema:
@copilot review — addressed both comments:

1. **Gemini L402**: extended the comptime gate from `if (!has_events) return;` to `if (!has_events or !has_hooks) return;`. Zero-hook builds now fold `emitSync` entirely away at comptime, matching `emitHook`'s shortcut.

2. **Copilot L400**: doc comment now calls out the two real hazards (re-entrancy, and ordering relative to buffered events) so plugin authors see the tradeoff before defaulting to sync dispatch.

Tests green (223/223).

Issue #286, opened 2026-02-13
Make zbgfx a lazy dependency (only fetch for bgfx backend)


apotema:
## Context

Similar to the wgpu_native_zig lazy dependency fix (#285), `zbgfx` is currently fetched unconditionally for all desktop builds in `build_helpers/deps_graphics.zig`, but it's only needed when `backend == .bgfx`.

## Proposed Change

Apply the same conditional loading pattern used for wgpu_native:

```zig
const zbgfx: ?*std.Build.Module = if (backend == .bgfx)
    labelle_dep.builder.dependency("zbgfx", .{
        .target = target,
        .optimize = optimize,
    }).module("zbgfx")
else
    null;
```

This would also require marking `zbgfx` as `.lazy = true` in labelle-gfx's `build.zig.zon`.

## References

- Follows the pattern established in #285
- Flagged by Copilot review on #285

Issue #12, opened 2025-12-08
Investigate example_3 render loop crash with atlas sprites


apotema:
## Description

Example_3 crashes in the render loop when running with `CI_TEST=1`. The crash occurs in `updatePosition` because sprites created from atlas data via `loadAtlasComptime` aren't properly registered in RetainedEngine's sprites map.

## Current Workaround

Example_3 exits early in CI mode after assertions pass:

```zig
// In CI mode, exit after assertions pass
// The render loop still has issues with sprite/atlas sync that need investigation
// (updatePosition crashes when sprites aren't properly created from atlas)
if (ci_test) {
    std.debug.print("CI mode: exiting after assertions\n", .{});
    return;
}
```

## Suspected Issue

The scene loading flow creates entities with Sprite components, but the sprites may not be properly created in the RetainedEngine when:
1. Atlas is loaded via `loadAtlasComptime`
2. Scene entities reference sprites from that atlas
3. `RenderPipeline.sync()` is called
4. `updatePosition()` fails because the sprite doesn't exist in the engine's internal map

## Steps to Reproduce

1. Run `CI_TEST=1 zig build example-3` (currently skips render loop)
2. Remove the early exit workaround
3. Observe crash in `updatePosition`

## Investigation Needed

- [ ] Verify sprites are being created in RetainedEngine when `RenderPipeline.sync()` calls `createSprite`
- [ ] Check if atlas sprites require different handling than regular sprites
- [ ] Determine if the issue is in labelle-engine or labelle-gfx
- [ ] Add proper error handling/logging to identify the exact failure point

## Related

- labelle-gfx v0.12.0 added `loadAtlasComptime` method
- PR #11 updated to use this method but had to keep the CI early exit

## Files

- `usage/example_3/main.zig` - Example with workaround
- `src/render_pipeline.zig` - RenderPipeline.sync() implementation
- `src/scene.zig` - Scene loading logic

PR #470, opened 2026-04-22
feat(save_load): persist Parent component alongside Position

apotema:
## Summary
The save mixin only writes / reads components that appear in the game's `ComponentRegistry`. `Parent` (i.e. `core.ParentComponent(Entity)`) isn't registered there — it's instantiated inside `GameConfig` and written by `setParent`. So even with [labelle-core 1.11.0](https://github.com/labelle-toolkit/labelle-core/releases/tag/v1.11.0) marking `ParentComponent` `.saveable`, the engine still never emitted Parent to the save file; children of parented prefab nodes (room decor, fridge shelves, canteen tables, hunger-carry items, …) lost their parent on load and their saved local Position rendered as world-space, drifting to scene origin.

Mirror the built-in Position pathway for Parent:
- **Save**: if the entity has a `Parent` component, emit a `"Parent": {\"entity\": ..., \"inherit_rotation\": ..., \"inherit_scale\": ...}` object.
- **Load**: after the registry-driven component restore, if the entity's JSON carries a Parent block, look the `entity` up in the load `id_map` and call `setParent` so the engine's `Children` back-link is rebuilt alongside.

## Why not register Parent in the game's ComponentRegistry?
`ParentComponent(Entity)` is parameterised by the ECS backend's `Entity` type, so the concrete Parent type only exists inside `GameConfig`. The registry is assembler-generated from game-authored files; it can't reach the engine-internal specialisation. Treating Parent as a built-in (like Position) is the right layer for this.

## Surfaces
- [labelle-core #11 / #12](https://github.com/labelle-toolkit/labelle-core/pull/12) – declared Parent saveable + extended `remapEntityRefs` to narrower unsigned-integer Entity types (zig-ecs uses `u32`).
- [flying-platform-labelle #286](https://github.com/Flying-Platform/flying-platform-labelle/issues/286) – the game-side save/load decor work that surfaced all of this.

## Test plan
- [x] `zig build test` — existing suite passes
- [ ] Downstream smoke in flying-platform-labelle on a build that bumps `engine_version` — F5 → F9 keeps the hydroponics / condenser backgrounds / pots / shelves at their centered positions instead of snapping to scene origin.

### DIFF
--- a/src/game/save_load_mixin.zig
+++ b/src/game/save_load_mixin.zig
@@ -292,26 +292,42 @@ pub fn Mixin(comptime Game: type) type {
                 // save block above. Uses `setParent` so the engine's
                 // Children back-link is rebuilt alongside.
                 //
+                // Mirrors the save-side `has_parent_in_registry` guard
+                // and Position's load-side guard: if a game ever
+                // registers `Game.ParentComp`, the registry-driven
+                // restore already handles Parent via `addComponent`, so
+                // this built-in block would double-restore and trigger
+                // `setParent` with stale state (review #470).
+                //
                 // Skip the restore unless we can both parse a valid
                 // saved entity ID and map it to a post-load entity —
                 // passing a stale or raw ID to `setParent` would trip
                 // `assertEntityAlive` in debug or corrupt hierarchy
-                // state in release builds (cursor / Copilot review).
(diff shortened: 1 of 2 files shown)


apotema:
@gemini review

PR #472, opened 2026-04-23
docs(rfc): prefab foundations — save/load + declarative animation

apotema:
## Summary

Two paired RFCs covering the "author in prefabs, let the engine handle round-trip and animation" story. Each builds on the other.

### [RFC-SAVE-LOAD-PREFABS.md](RFC-SAVE-LOAD-PREFABS.md)
Prefab-aware save/load. Record each entity's prefab source + instance overrides on save; on load re-instantiate the prefab (Phase 1 — structure) then apply saved overrides (Phase 2 — state). Non-prefab entities stay on the current flat save/load path.

Deletes the per-entity re-hydration scripts games hand-roll today:
- flying-platform-labelle's `RoomDecor` marker + switch table (born yesterday from #286)
- `HydroponicsPlant.saveable` + the Sprite re-add block in its animation
- `restoreSprites` in the game's save_load.zig
- The `needsReinit` / `initOverlays` dances in the condenser + kitchen animation scripts (six iterations of hardening on #286 traces back to stale-ID recovery that this RFC eliminates)

Built on #467 (postLoad on nested entities) and #470 (Parent persistence).

### [RFC-PREFAB-ANIMATION.md](RFC-PREFAB-ANIMATION.md)
Two engine components authored declaratively in prefabs:
- `SpriteAnimation { frames, fps, mode }` — frame cycling (loop / once / ping_pong).
- `SpriteByField { component, field, source, entries }` — pick sprite from a map keyed by a runtime field value (level → sprite for hydroponics plants, etc.).

Engine tick systems walk `view(.{SpriteAnimation, Sprite}, .{})` / `view(.{SpriteByField, Sprite}, .{})` and mutate `Sprite`. Runtime add/remove is the gate — engine stays ignorant of game predicates.

Net after adoption: ~600 lines of game-side animation code deleted, replaced by two engine components + two tick systems.

Complements the existing `AnimationDef` / `AnimationState` character animation; this is for the single-clip simple case that was previously hand-rolled.

## Why together

The two RFCs are conceptually paired:
- Save/load-for-prefabs gives us "non-saveable components come back on load."
- Prefab-animation gives us "declarative per-tick mutation with in-flight state excluded from save."

Together they eliminate the re-hydration + manual ticking that every room's decor, every workstation's overlay, and every character animation's Sprite restoration has been reinventing.

## Status

**Draft.** Seeking review on:
- Two-phase load structure + `PrefabInstance` / `PrefabChild` built-in markers (save/load RFC §Architecture).
[...]

PR #474, opened 2026-04-23
feat(save_load): PrefabInstance + PrefabChild built-in handlers (Slice 1b)

apotema:
## Summary

Re-opened from #473 (which auto-closed when its base `feat/save-parent-component` was deleted on merge). This is the same content, rebased onto `main` post-#470 and onto labelle-core v1.12.0. Slice 1b of the save/load-for-prefabs design in RFC #472.

Serialises the two prefab-lineage components from [labelle-core v1.12.0](https://github.com/labelle-toolkit/labelle-core/pull/13) — `PrefabInstance { path, overrides }` and `PrefabChild(Entity) { root, local_path }` — as built-ins in the save mixin, same channel `Position` and `Parent` use.

## What's in it

- `src/game.zig` — expose `PrefabInstanceComp` + `PrefabChildComp` on `GameConfig` (mirrors `ParentComp` / `ChildrenComp`).
- `src/game/save_load_mixin.zig`:
  - **New `isRegistered(T)` comptime helper** — extracts the repeated `for (names) ... break :blk` boilerplate from Position, Parent, PrefabInstance, PrefabChild guard sites into a single helper (−20 lines of duplication).
  - **`writeJsonString` helper** — RFC 8259-compliant string escape for `"`, `\`, and control chars (short escapes for `\b \t \n \f \r`, `\uXXXX` for the rest).
  - Save-side blocks for PrefabInstance + PrefabChild after the Parent block. Guarded by `isRegistered` type-identity check so a game registering either type directly doesn't produce duplicate JSON keys.
  - Load-side blocks. Defensive `.object` + field-type switches. String fields duped into `active_world.nested_entity_arena` to outlive the parsed JSON's `deinit`. `PrefabChild.root` remapped via `id_map`.
- `test/save_load_mixin_test.zig` — **4 new tests**:
  1. Round-trip: path + overrides survive as exact strings (with escape regression guard on `\"components\"`), `PrefabChild.root` remapped correctly.
  2. Malformed `PrefabInstance` JSON value is skipped, not panicked (`switch (pi_val) { .object => ..., else => break :blk }` regression guard).
  3. Unresolvable `PrefabChild.root` is skipped, not attached to stale id (`id_map.get orelse break :blk` regression guard).
  4. Control-char round-trip — `overrides` with `\n`, `\t`, `\\`, `\"`, `\r`, `\b`, `\f`, and a raw control char all survive byte-for-byte.

## What this does NOT do

Explicitly scoped to data-plumbing. Later slices:

- **Slice 2** — `spawnFromPrefab` API on `Game` + jsonc bridge auto-tagging of prefab-sourced entities.
- **Slice 3** — two-phase load (Phase 1 reinstantiates prefabs before Phase 2 applies overrides).
- **Slice 4** — downstream migration (delete `RoomDecor` etc. in flying-platform-labelle).

Entity collection on save still iterates registered components only, so an entity carrying ONLY built-in components (no registered ones) still won't appear in the save file. Same limitation as `Parent` today.

## Dependencies

- **Depends on** labelle-core v1.12.0 (`PrefabInstance` + `PrefabChild` types) — already released.
- **Builds on** #470 (`feat(save_load): persist Parent component`) — merged, so CI clones the correct main.

## Test plan

- [x] `zig build test` — full engine suite green (was 204 tests pre-slice-1; now 208 after this PR's 4 new tests).
- [x] Round-trip verified including escape / malformed / id_map-miss / control-char paths.
- [ ] Downstream smoke in flying-platform-labelle — waits on the pin-bump PR to v1.23.0 / core 1.12.0 (coming next).

Previous conversation context on #473: reviews from gemini were addressed in prior force-pushes (bounds check, isRegistered helper refactor now applied to all 4 built-ins, defensive path tests).

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -9,6 +9,8 @@ const ComponentPayload = hooks_types.ComponentPayload;
 const VisualType = core.VisualType;
 const ParentComponent = core.ParentComponent;
 const ChildrenComponent = core.ChildrenComponent;
+const PrefabInstance = core.PrefabInstance;
+const PrefabChild = core.PrefabChild;
 
 const atlas_mod = @import("atlas.zig");
 const assets_mod = @import("assets/mod.zig");
@@ -53,6 +55,7 @@ pub fn GameConfig(
     const Icon = if (@hasDecl(RenderImpl, "Icon")) RenderImpl.Icon else void;
     const Parent = ParentComponent(Entity);
     const Children = ChildrenComponent(Entity);
+    const PrefabChildT = PrefabChild(Entity);
     const EnginePayload = hooks_types.HookPayload(Entity);
     const has_events = GameEvents != void;
     const Payload = if (has_events) core.MergeHookPayloads(.{ EnginePayload, GameEvents }) else EnginePayload;
@@ -82,6 +85,8 @@ pub fn GameConfig(
         pub const IconComp = Icon;
         pub const ParentComp = Parent;
         pub const ChildrenComp = Children;
+        pub const PrefabInstanceComp = PrefabInstance;
+        pub const PrefabChildComp = PrefabChildT;
         pub const Input = @import("input.zig").InputInterface(InputImpl);
         pub const Audio = @import("audio.zig").AudioInterface(AudioImpl);
         pub const Gui = @import("gui.zig").GuiInterface(GuiImpl);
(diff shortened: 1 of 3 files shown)


apotema:
@copilot review

PR #475, opened 2026-04-23
feat(animation): SpriteAnimation — declarative frame-cycling component (Phase A)

apotema:
## Summary

Phase A of [RFC-PREFAB-ANIMATION.md](https://github.com/labelle-toolkit/labelle-engine/pull/472) — a lightweight per-frame sprite cycling component for the "one clip, fixed frames, cycle at N fps" case currently hand-rolled in each game (condenser pipe, kitchen smoke, hydroponics overlay — ~600 lines of near-copies across three scripts in flying-platform-labelle alone).

Complements the existing `AnimationState` + `AnimationDef` system, which is tuned for characters with many clips × many variants × precomputed sprite tables. `SpriteAnimation` is the simpler primitive: no variants, no transitions, no comptime tables — just a frame array + fps + mode.

## Scope — pure state machine only

This PR adds the **component type + `advance(dt)` method**. Not included:

- ECS tick system walking `view(.{SpriteAnimation, Sprite}, .{})` and mutating the Sprite's `sprite_name`/`source_rect`/`texture` on frame flips. Lands in a follow-up — depends on game-backend atlas lookup (`game.findSprite`) and is easier to design once the state machine is settled.
- `SpriteByField` component from the same RFC (Phase B).
- Downstream migration of `condenser_animation.zig` etc. (Phase C, depends on save/load-for-prefabs for `spawnFromPrefab`).

Each independently shippable.

## Component shape

```zig
pub const SpriteAnimation = struct {
    pub const save = Saveable(.saveable, @This(), .{
        .skip = &.{ "timer", "frame", "forward" },
    });
    frames: []const []const u8,
    fps: f32,
    mode: AnimationMode = .loop,   // .loop / .once / .ping_pong
    // runtime state, skipped from save
    timer: f32 = 0,
    frame: u8 = 0,
    forward: bool = true,           // ping_pong direction
};
```

## Modes

- **`.loop`** — `frame = (frame + steps) % frames.len`.
- **`.once`** — plays through, stops on the last frame. Game removes the component to replay.
- **`.ping_pong`** — walks forward to `frames.len - 1`, then reverse to 0, flipping direction at each endpoint. Iterates per-step so the `forward` flag reflects "direction we JUST moved in" — flipping strictly on reversal, not on landing at an endpoint. (A closed-form modular version in the first draft hit the ambiguity at frames 0 and `len-1`; the comment in `src/sprite_animation.zig` explains the tradeoff.)

## Save policy

`.saveable` with `timer` / `frame` / `forward` skipped via `Saveable.skip`. On load, the animation starts from frame 0 — one-frame visual continuity loss is invisible at 60 fps, and save files stay small/deterministic. Explicit design call documented in the RFC.

## Tests

11 new tests in `test/sprite_animation_test.zig`:

- `.loop` — cycle, idempotent within a frame, large-dt multiple frames in one call.
- `.once` — stops on last frame, no further advance.
- `.ping_pong` — forward → peak → reverse → trough → forward sequence with forward-flag checks at each endpoint, single-frame stationary case.
- Degenerate inputs — zero frames, zero fps — both no-op cleanly.
- Save-policy contract — `.saveable`, with `timer`/`frame`/`forward` in the skip list (regression guard: if someone removes `.skip`, save files start bloating with per-tick timer drift).
- `currentSprite` — happy path + empty-frames null case.

Full engine suite: **214/214 green** (was 202 pre-slice-1; +11 new + misc recount from other recently-landed tests).

## Relationship to existing animation

| Component | Use case | Frames | Clips |
|---|---|---|---|
| `AnimationDef(zon)` / `AnimationState` | Characters | Per-clip, precomputed sprite table | Many (walk/idle/carry) × variants |
| `SpriteAnimation` | Simple overlay cycles | Inline `[]const []const u8` | One |

Games will use `AnimationState` for workers and `SpriteAnimation` for everything else. Neither obsoletes the other.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -45,6 +45,7 @@ pub fn build(b: *std.Build) void {
         "test/asset_catalog_test.zig",
         "test/asset_streaming_shim_test.zig",
         "test/animation_def_test.zig",
+        "test/sprite_animation_test.zig",
         "test/scene_assets_hooks_test.zig",
         "test/pause_hook_test.zig",
     };
--- a/src/root.zig
+++ b/src/root.zig
@@ -19,6 +19,7 @@ pub const hooks_types_mod = @import("hooks_types.zig");
 pub const animation_mod = @import("animation.zig");
 pub const animation_def_mod = @import("animation_def.zig");
 pub const animation_state_mod = @import("animation_state.zig");
+pub const sprite_animation_mod = @import("sprite_animation.zig");
 pub const atlas_mod = @import("atlas.zig");
 pub const assets_mod = @import("assets/mod.zig");
 pub const jsonc_mod = @import("jsonc");
@@ -124,6 +125,8 @@ pub const AnimationDef = animation_def_mod.AnimationDef;
 pub const AnimationState = animation_state_mod.AnimationState;
 pub const AnimMode = animation_def_mod.Mode;
 pub const AnimClipMeta = animation_def_mod.ClipMeta;
+pub const SpriteAnimation = sprite_animation_mod.SpriteAnimation;
+pub const SpriteAnimationMode = sprite_animation_mod.AnimationMode;
 
 // ── Atlas ──
 pub const SpriteData = atlas_mod.SpriteData;
--- a/src/sprite_animation.zig
+++ b/src/sprite_animation.zig
@@ -1,0 +1,153 @@
+//! SpriteAnimation — declarative per-frame sprite cycling.
+//!
+//! For simple animations that just walk an array of sprite names at a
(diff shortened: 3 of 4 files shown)


apotema:
@copilot review

apotema:
@copilot review — addressed gemini HIGH (negative-timer → u32 cast panic) in 01b2f28: clamp `timer` to `>= 0` after `+= dt` so neither a negative dt nor floating-point residual from the previous subtraction can drive `steps_f` negative before the `@intFromFloat(@floor(...))` → u32 cast. Two regression tests cover both paths.

PR #476, opened 2026-04-23
feat(animation): SpriteByField — field-driven sprite selection (Phase B)

apotema:
## Summary

Phase B of [RFC-PREFAB-ANIMATION.md](https://github.com/labelle-toolkit/labelle-engine/pull/472), sibling to [#475](https://github.com/labelle-toolkit/labelle-engine/pull/475) (Phase A — `SpriteAnimation`). Where `SpriteAnimation` handles the time-driven frame cycle case, `SpriteByField` handles the field-driven case: the sprite follows the runtime value of a named field on another component.

Canonical example: the hydroponics plant overlay swaps sprite based on `TendableWorkstation.level` — 0/1 hidden, 2 → sapling_lvl1, 3 → sapling_lvl2, 4/5 → green. Today that's ~140 lines in `hydroponics_animation.zig`; once the tick system lands (Phase B+), this component collapses the logic to prefab data.

## Component shape

```zig
pub const SpriteByField = struct {
    pub const save = Saveable(.saveable, @This(), .{
        .skip = &.{ "last_key_set", "last_key" },
    });

    pub const Entry = struct {
        key: i32,                    // signed (gemini review on #472)
        sprite_name: ?[]const u8,    // null → hide
    };

    component: []const u8,
    field: []const u8,
    source: SpriteByFieldSource = .self,  // .self or .parent
    entries: []const Entry,

    last_key_set: bool = false,      // tick cache, skipped from save
    last_key: i32 = 0,
};
```

## Scope — pure state machine

This PR adds the component type + `lookup(key: i32)` method that returns `.match(?[]const u8)` or `.no_match`. The ECS tick system that reads the runtime field value, coerces to `i32`, calls `lookup`, and mutates the Sprite on match lands in a follow-up — depends on `ComponentRegistry.getType` + `std.meta.fieldIndex` reflection + `.parent` traversal, all easier to design once the data shape is settled.

Same staging pattern as [#475](https://github.com/labelle-toolkit/labelle-engine/pull/475) (Phase A state machine).

## `lookup` result

Union for three distinguished cases:
- **`.match`** with non-null sprite — normal swap.
- **`.match`** with null sprite — "hide" (tick system sets `visible = false`).
- **`.no_match`** — no entry matched, tick system leaves Sprite alone rather than guessing.

The `match` / `no_match` distinction matters because a table can legitimately contain a null entry (e.g. level 0 hides the overlay), and the tick system needs to distinguish "no entry for this value" from "entry says hide."

## Save policy

`.saveable` with `last_key_set` / `last_key` runtime cache skipped via `Saveable.skip`. On load, the cache resets and the next tick re-resolves the whole pipeline — negligible cost, and avoids the failure mode where a stale cache masks a legitimate sprite update on the first post-load tick.

## Tests (8 new in `test/sprite_by_field_test.zig`)

- Lookup resolves each configured key in the canonical 6-row hydroponics table.
- Null `sprite_name` means "hide" (distinct from `.no_match`).
- Unmapped keys return `.no_match`.
- Signed keys supported (sentinel `-1` works — the gemini-reviewed change from `u32` to `i32`).
- Empty entries table returns `.no_match` for every key.
- First-match-wins on duplicate keys (prefab-authoring bug, but the rule is pinned).
- Save-policy regression guard: `last_key_set` / `last_key` are in the skip list.
- Default `source` is `.self` — silent flip to `.parent` would break downstream games.

Full engine suite: **211/211 green** (+8 new).

## Relationship to Phase A (#475)

Both PRs are pure state-machine slices of the same RFC; independent of each other and independent of everything else. Merging order doesn't matter — both land their tick systems as separate follow-up PRs.

## Parallel thread

RFC progress on engine side:

| Phase | PR | State |
|---|---|---|
| A — SpriteAnimation | #475 | Draft, under review |
| B — SpriteByField | **this PR** | **Draft, new** |
| A+ — SpriteAnimation tick | — | Not started |
| B+ — SpriteByField tick | — | Not started |
| C — flying-platform migration | — | Blocked on tick systems + save/load Slice 2 |

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -46,6 +46,7 @@ pub fn build(b: *std.Build) void {
         "test/asset_streaming_shim_test.zig",
         "test/animation_def_test.zig",
         "test/sprite_animation_test.zig",
+        "test/sprite_by_field_test.zig",
         "test/scene_assets_hooks_test.zig",
         "test/pause_hook_test.zig",
         "test/spawn_from_prefab_test.zig",
--- a/src/root.zig
+++ b/src/root.zig
@@ -20,6 +20,7 @@ pub const animation_mod = @import("animation.zig");
 pub const animation_def_mod = @import("animation_def.zig");
 pub const animation_state_mod = @import("animation_state.zig");
 pub const sprite_animation_mod = @import("sprite_animation.zig");
+pub const sprite_by_field_mod = @import("sprite_by_field.zig");
 pub const atlas_mod = @import("atlas.zig");
 pub const assets_mod = @import("assets/mod.zig");
 pub const jsonc_mod = @import("jsonc");
@@ -127,6 +128,8 @@ pub const AnimMode = animation_def_mod.Mode;
 pub const AnimClipMeta = animation_def_mod.ClipMeta;
 pub const SpriteAnimation = sprite_animation_mod.SpriteAnimation;
 pub const SpriteAnimationMode = sprite_animation_mod.AnimationMode;
+pub const SpriteByField = sprite_by_field_mod.SpriteByField;
+pub const SpriteByFieldSource = sprite_by_field_mod.SpriteByFieldSource;
 
 // ── Atlas ──
 pub const SpriteData = atlas_mod.SpriteData;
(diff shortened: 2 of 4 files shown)


apotema:
@copilot review

PR #483, opened 2026-04-23
feat(scene-bridge): auto-tag prefab entities on scene load (Slice 2b)

apotema:
## Summary

Slice 2b of [RFC-SAVE-LOAD-PREFABS.md](https://github.com/labelle-toolkit/labelle-engine/pull/472). Scene-load counterpart to [#482](https://github.com/labelle-toolkit/labelle-engine/pull/482) (Slice 2a runtime `spawnFromPrefab`) — entities whose scene jsonc declared a `"prefab"` field now get tagged with `PrefabInstance` on load. Games don't have to call `spawnFromPrefab` manually for scene-declared prefab instances.

## Change

One insertion in `src/jsonc_scene_bridge.zig::loadEntityInternal` at the end of entity assembly (after components applied, children spawned, `onReady` fired): if the scene jsonc had `"prefab": "..."`, dupe the path into `active_world.nested_entity_arena` and `addComponent` `PrefabInstance { path, overrides: "" }`.

## Scope — root only

Explicitly tagging the root entity only, NOT nested children. The children loop in `loadEntityInternal` interleaves prefab-defined children and scene-declared children and recurses through `loadChildEntity`; picking out which children came from the prefab and generating `children[i]...` paths matching the runtime `spawnFromPrefab` format ([#482](https://github.com/labelle-toolkit/labelle-engine/pull/482)) is a separate design problem worth its own PR.

- Runtime `spawnFromPrefab` ([#482](https://github.com/labelle-toolkit/labelle-engine/pull/482)) tags root + all descendants with `children[i]...`.
- Scene-load (this PR) tags root only — sufficient for the save mixin to recognise prefab-sourced entities; not yet sufficient for full Slice 3 two-phase load of nested-prefab scenes.

## Tests (4 new in `test/jsonc_bridge_prefab_tags_test.zig`)

1. Prefab-sourced scene entity → `PrefabInstance { path: "enemy" }`.
2. Non-prefab scene entity → no `PrefabInstance`.
3. Multiple prefab instances — 2× warrior + 1× archer each get their own tag with the correct path.
4. **Save/load round-trip** — scene-load tags → `saveGameState` → `resetEcsBackend` → `loadGameState` → tags restored. Pins the contract with Slice 1b's built-in save handler via a byte-level check on the save file (`"PrefabInstance": ...` + `"path": "unit"`).

### Gotcha caught during test development

[...]

PR #482, opened 2026-04-23
feat(prefab): spawnFromPrefab runtime API with lineage tags (Slice 2a)

apotema:
## Summary

Slice 2 of [RFC-SAVE-LOAD-PREFABS.md](https://github.com/labelle-toolkit/labelle-engine/pull/472). Stacked on [#474](https://github.com/labelle-toolkit/labelle-engine/pull/474) (Slice 1b built-in handlers).

New `game.spawnFromPrefab(path, pos) ?Entity` — wraps the existing `game.spawnPrefab` with automatic `PrefabInstance` / `PrefabChild` tagging so the save mixin's built-in handlers can round-trip the prefab lineage through F5 → F9.

Closes the `spawnFromPrefab` half of [#479](https://github.com/labelle-toolkit/labelle-engine/issues/479). The jsonc-bridge auto-tagging half is the natural follow-up.

## What's in it

1. **`Game.spawnFromPrefab(path, pos)`** — delegates to the existing `spawnPrefab` hook (already wired via `JsoncSceneBridge::loadSceneFromSource`), then tags:
   - Root → `PrefabInstance { path: <arena-duped>, overrides: "" }`.
   - Each descendant → `PrefabChild { root, local_path }` where `local_path` is a dotted-index path (`children[0]`, `children[1].children[0]`, etc.).
   - String storage in `active_world.nested_entity_arena` — same ownership contract [labelle-core v1.12](https://github.com/labelle-toolkit/labelle-core/pull/13) documented for the tracking types (arena-owned, freed on scene change, no per-entity free needed).

2. **Recursive child tagger** — `tagPrefabChildren(root, parent, base_path, arena)` walks `game.getChildren(parent)` and emits `children[i]` → `children[i].children[j]` paths as it descends.

## What this does NOT do yet

Explicitly scoped. Follow-ups:

- **Structured overrides** — `overrides = ""` on every spawn. The full "apply caller-supplied component overrides on top of prefab defaults" pipeline needs design work for `serde.writeComponent` round-trip + application semantics.
- **jsonc-bridge auto-tagging** (the other half of #479) — scene-sourced prefab entities don't get tagged yet. Lands as a separate PR once the runtime API shape settles here.
- **Two-phase load** (Slice 3) — Phase 1 reinstantiating from saved `PrefabInstance` + matching saved `PrefabChild` records to newly-spawned children via `(root, local_path)`. This slice establishes the `local_path` format Slice 3 will match against.

[...]

PR #481, opened 2026-04-23
feat(animation): SpriteByField ECS tick system (Phase B+, closes #478)

apotema:
## Summary

Phase B+ of [RFC-PREFAB-ANIMATION.md](https://github.com/labelle-toolkit/labelle-engine/pull/472). Stacked on top of [#476](https://github.com/labelle-toolkit/labelle-engine/pull/476) (Phase B — pure state machine). Adds the ECS tick layer that walks entities carrying `SpriteByField` + Sprite, resolves the driving component + field via the game's `ComponentRegistry`, and applies the lookup result to the Sprite.

Closes [#478](https://github.com/labelle-toolkit/labelle-engine/issues/478).

## The tricky bit — runtime name resolution

`SpriteByField.component` and `.field` are **runtime** strings (prefab-author-facing). Zig's comptime wants comptime-known names for type and field lookup. The tick uses the "comptime iterate, runtime match" pattern:

```zig
inline for (comptime Reg.names()) |comp_name| {
    if (std.mem.eql(u8, comp_name, sbf.component)) {
        const T = Reg.getType(comp_name);           // comptime in this branch
        const comp = ecs.getComponent(target, T);
        // Field lookup is the same shape on @typeInfo(T).@"struct".fields.
    }
}
```

Same codegen shape `save_load_mixin`'s `inline for (names)` loop already uses. Unknown component / unknown field → silent skip that entity this tick, matching the plugin-controller duck-typing contract (`production`, `command_buffer`).

## Field coercion to `i32`

Integer fields (signed or unsigned, any width) clamp-widen via `i64` intermediate. Enum fields go through `@intFromEnum`. Unsupported field types (floats, slices, nested structs) return null from the field reader and the tick skips. Validation at spawn time will be added once `spawnFromPrefab` exists ([Slice 2 / #479](https://github.com/labelle-toolkit/labelle-engine/issues/479)).

## Cache

`last_key_set` + `last_key` short-circuit the whole pipeline when the driving field value hasn't changed since last tick. Idle entities write nothing — matches the contract [Phase A+ (#480)](https://github.com/labelle-toolkit/labelle-engine/pull/480) established for `SpriteAnimation`.

## Tests (10 new in `test/sprite_by_field_tick_test.zig`)

1. Integer field → sprite_name (happy path).
2. Enum field → sprite_name via `@intFromEnum`.
3. `.parent` source reads driving field off parent entity (via `game.getParent`).
[...]

PR #480, opened 2026-04-23
feat(animation): SpriteAnimation ECS tick system (Phase A+, closes #477)

apotema:
## Summary

Phase A+ of [RFC-PREFAB-ANIMATION.md](https://github.com/labelle-toolkit/labelle-engine/pull/472). Stacked on top of [#475](https://github.com/labelle-toolkit/labelle-engine/pull/475) (Phase A — pure state machine). Adds the ECS layer that walks entities carrying both `SpriteAnimation` + Sprite and mutates the Sprite on frame flips.

Closes [#477](https://github.com/labelle-toolkit/labelle-engine/issues/477).

## What's in it

- `src/sprite_animation_tick.zig` — single `tick(game, dt)` free function. Generic over the game type, same shape `save_load_mixin` uses. Games call it once per frame from their scene-tick slot (assembler auto-wire can come later).
- Atlas resolution is guarded by `@hasField` so the tick compiles against `StubRender.Sprite` (engine-level tests, no atlas fields) as well as the full labelle-gfx Sprite. Real games always hit the resolution branch; stub builds skip it cleanly.
- `markVisualDirty(entity)` fires on frame flip only. Idle ticks (sub-frame `dt`) write nothing and produce no dirty signal.

## Semantics locked by tests

The 6 new tests in `test/sprite_animation_tick_test.zig` pin:

1. Frame-flip tick writes new `sprite_name`.
2. **Idle (sub-frame) tick does NOT overwrite** — regression guard against a naive "always sync on tick" implementation that would clobber an entity's pre-existing `sprite_name` with frame 0 on the first tick.
3. Entity without SpriteAnimation is skipped (untouched).
4. Entity with SpriteAnimation but no Sprite is skipped — view filter regression guard.
5. `.once` mode stops writing after the cycle completes (even large-dt ticks don't restart).
6. Multiple entities advance independently at different fps.

Full engine suite: **220/220 green** (+6 new).

## What this does NOT do

- No equivalent for `SpriteByField` (#476) — that tick is its own follow-up ([#478](https://github.com/labelle-toolkit/labelle-engine/issues/478)), with the extra complexity of field reflection + `.parent` traversal.
- No assembler auto-wiring — games call `engine.spriteAnimationTick(&game, dt)` from their own scene-tick slot. A later slice can auto-wire via the assembler.
[...]

PR #484, opened 2026-04-23
feat(save_load): two-phase load — prefab re-instantiation on F9 (Slice 3)

apotema:
## Summary

**Slice 3 — the payoff slice.** Closes the RFC's core promise: prefab-sourced entities come back from `loadGameState` via `spawnFromPrefab`, bringing their non-saveable components (Sprite, animation overlays, etc.) *and* child structure along for free. Saved component values apply on top as overrides.

Once this lands, flying-platform-labelle's `RoomDecor` marker + `restoreSprites` switch table + `HydroponicsPlant.saveable` + `condenser_animation::needsReinit` are all deletable. That's the migration milestone (Phase C).

## The new load flow

```
loadGameState:
  Step 1 — resetEcsBackend (unchanged)
  Step 2 — build id_map (NEW: three-pass)
    Phase 1a: for each saved PrefabInstance:
      new_root = spawnFromPrefab(path, saved_Position)
      id_map[saved_root_id] = new_root
    Phase 1b: for each saved PrefabChild { root, local_path }:
      new_child = walk(id_map[root], local_path)   // via getChildren
      id_map[saved_child_id] = new_child
    Phase 1c: for saved entities not yet in id_map:
      new = createEntity()                          // v2 fallback
      id_map[saved_id] = new
  Step 3 — apply components (UNCHANGED)
```

Phase 2 (the component-restoration loop) is untouched — it just now operates on entities already-populated by Phase 1. Saved component data applies on top as overrides. Non-prefab entities still hit the v2 path via Phase 1c's fallback.

## `findChildByLocalPath` helper

New private helper that parses a `children[i].children[j]...` path (the format `spawnFromPrefab` / the scene-bridge auto-tagger emit) and walks `game.getChildren` step by step. Returns `null` on parse error or missing index — Phase 1b logs a warning and skips rather than asserts, so malformed saves can't panic the load.

## Tests (4 new in `test/save_load_two_phase_test.zig`)

1. **End-to-end round-trip** — spawn prefab with child, mutate Health on both, save, reset, load. Asserts the root's PrefabInstance tag, the mutated Health on root + child, the child came back via `getChildren` (not a fresh blank), and the child still carries `PrefabChild`.
2. **No duplicate entities** — a 3-child prefab loads as exactly 4 entities total (1 root + 3 children), not 7 (4 + 3 ghosts). Regression guard for the Phase 1c fallback firing on PrefabChild entries.
3. **Non-prefab entities still work** — mixed save with 1 prefab-sourced + 1 plain entity round-trips both.
4. **Graceful fallback on missing prefab** — hand-edits a save to reference a non-existent prefab, asserts the entity comes back as a fresh-created fallback. Guards the "renamed prefab between save and load" scenario.

Full engine suite: **219/219 green** (+4 new, and all pre-existing tests still pass against the new Phase 1 path).

## What this does NOT do yet

- **Structured overrides** — `PrefabInstance.overrides` is still `""` on spawn. The call site in Slice 3 reads `Position` from the saved components as a fixed arg to `spawnFromPrefab`. Full "apply caller-supplied component overrides" lands in a follow-up; today's Phase 2 step applies saved component values on top of prefab defaults, which covers the same ground for everything except "save-time values that the prefab prescribes."
- **Scene-bridge child tagging** (deferred from Slice 2b) — scene-loaded prefab entities don't yet get `PrefabChild` tags on their descendants, so scene-only saves lose descendant lineage. Runtime-spawned prefabs (via `spawnFromPrefab`) are fully tagged and round-trip correctly — this slice's tests exercise that path.

## Dependencies

Stacked chain: [#474](https://github.com/labelle-toolkit/labelle-engine/pull/474) (Slice 1b) → [#482](https://github.com/labelle-toolkit/labelle-engine/pull/482) (Slice 2a) → [#483](https://github.com/labelle-toolkit/labelle-engine/pull/483) (Slice 2b) → **this PR**. Merge in order; retarget base as each parent merges.

## RFC end-to-end path

The RFC's migration examples are now achievable after this PR:

- flying-platform-labelle `RoomDecor` marker + `restoreSprites` switch table → **deletable**.
- `HydroponicsPlant.saveable` marker → **`.transient`** (prefab re-instantiation covers it).
- `condenser_animation::needsReinit` / `initOverlays` dance → **deletable** once overlays move into prefabs + `SpriteAnimation` ([#480](https://github.com/labelle-toolkit/labelle-engine/pull/480) / [#481](https://github.com/labelle-toolkit/labelle-engine/pull/481)).

Phase C migration is the next milestone.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -49,6 +49,7 @@ pub fn build(b: *std.Build) void {
         "test/pause_hook_test.zig",
         "test/spawn_from_prefab_test.zig",
         "test/jsonc_bridge_prefab_tags_test.zig",
+        "test/save_load_two_phase_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/game/save_load_mixin.zig
+++ b/src/game/save_load_mixin.zig
@@ -48,6 +48,32 @@ pub fn Mixin(comptime Game: type) type {
             };
         }
 
+        /// Walk a dotted `children[i]...` path from `root` through
+        /// `game.getChildren`, returning the entity at the end or
+        /// `null` when the path doesn't resolve (missing index,
+        /// malformed syntax, root has no children, etc.). Matches the
+        /// path format `spawnFromPrefab` (and the scene-bridge
+        /// auto-tagger) emit, so save/load Phase 1 can find the
+        /// re-spawned child that corresponds to each saved PrefabChild.
+        fn findChildByLocalPath(self: *Game, root: Entity, local_path: []const u8) ?Entity {
+            var current: Entity = root;
+            var rest = local_path;
+            while (rest.len > 0) {
+                if (rest[0] == '.') rest = rest[1..];
+                const prefix = "children[";
+                if (!std.mem.startsWith(u8, rest, prefix)) return null;
+                rest = rest[prefix.len..];
+                const close = std.mem.indexOfScalar(u8, rest, ']') orelse return null;
+                const idx = std.fmt.parseInt(usize, rest[0..close], 10) catch return null;
+                rest = rest[close + 1 ..];
+
+                const children = self.getChildren(current);
+                if (idx >= children.len) return null;
+                current = children[idx];
+            }
+            return current;
+        }
+
         /// Write a JSON-escaped string literal (including surrounding
         /// quotes) to `writer`. Used by the built-in save pathway for
         /// components with `[]const u8` fields (PrefabInstance.path,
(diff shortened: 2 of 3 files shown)


apotema:
@copilot review

apotema:
@copilot review — addressed the three inline comments plus a gap 1ffadc6 left in Step 3:

1. **L75 `findChildByLocalPath`**: empty-path → return null (explicit guard at top) so malformed `PrefabChild.local_path == ""` no longer aliases the child ID onto the root in `id_map`.
2. **L395 Phase 1a order**: `getSavedId` now runs before `spawnFromPrefab` — malformed entries skip without leaking a spawned tree.
3. **L391 scene_entities staleness**: added `self.scene_entities.clearRetainingCapacity()` in Step 1 so `resetEcsBackend` + `trackSceneEntity` via `spawnPrefabImpl` can't interleave stale and new IDs.
4. **Bonus**: Step 3 still had raw `.object` / `.integer` tag casts (the 1ffadc6 message claimed otherwise) — routed all four through the defensive accessors to match Phase 1.

Tests green (222/222).

PR #485, opened 2026-04-23
feat(scene-bridge): tag prefab descendants on scene load (Slice 2b+)

apotema:
## Summary

Completes the deferred piece of [Slice 2b (#483)](https://github.com/labelle-toolkit/labelle-engine/pull/483): scene-loaded prefab entities now get both `PrefabInstance` on the root **and** `PrefabChild` on every descendant, with the same `local_path` format [`spawnFromPrefab` (#482)](https://github.com/labelle-toolkit/labelle-engine/pull/482) emits. The [two-phase load (#484)](https://github.com/labelle-toolkit/labelle-engine/pull/484) can now match saved child IDs to respawned ones uniformly across scene-declared and runtime-spawned prefabs.

**With this slice, the save/load-for-prefabs RFC chain is feature-complete for downstream migration.**

## Refactor

Extracted the root+descendants tagging from `Game.spawnFromPrefab` into a public `Game.tagAsPrefabInstance(entity, path)`. Both paths call it:

- `Game.spawnFromPrefab` — create entity tree via `spawnPrefab`, then `tagAsPrefabInstance`.
- `JsoncSceneBridge.loadEntityInternal` — populate entity tree normally, then `tagAsPrefabInstance` at the end when the entity came from a scene-declared `"prefab"` field.

Running `tagAsPrefabInstance` **after** children are fully attached means `getChildren` walks the real tree, so `local_path`s generate cleanly in a single pass. No need to thread prefab-root state through the recursion.

## Tests (2 new)

1. **`jsonc_bridge_prefab_tags_test`** — scene-load of a tree prefab (2 children, one with a grandchild). Asserts root has `PrefabInstance`, the 3 descendants have `PrefabChild` with the expected `children[0]` / `children[1]` / `children[1].children[0]` paths, all pointing at the same root. Mirrors the runtime-spawn test in `spawn_from_prefab_test` — pinning the contract that both paths produce identical tags.

2. **`save_load_two_phase_test`** — the flying-platform-labelle scenario end-to-end:
   - Scene jsonc declares `{ "prefab": "room" }`.
   - Scene-load tags automatically.
   - Mutate child Health.
   - Save → reset → load.
   - Asserts: 1 `PrefabInstance` root; 2 children via `getChildren` (not duplicates, not stale); both children carry mutated Health (Phase 2 overrides); both carry `PrefabChild` pointing at the new root; total Health-carrying entities = 3 (no Phase 1c fallback ghosts).

Full engine suite: **221/221 green** (+2 new).

[...]

PR #488, opened 2026-04-23
feat+docs: end-to-end prefab+animation walkthrough + JSONC bridge slice/optional deserialization

### LINKED ISSUE #487: JSONC bridge: deserializer rejects slice types (blocks SpriteAnimation + SpriteByField in prefabs)
## Summary

`src/jsonc_scene_bridge.zig::deserialize` handles primitives, `[]const u8`, enums, tagged unions, and structs — but falls through to `null` for anything else, notably slice types like `[]const []const u8` or `[]const SomeStruct`.

Concrete consequence: the prefab-animation components introduced in #475 / #476 can't be declared inline in a prefab jsonc, even though the RFC (`RFC-PREFAB-ANIMATION.md`) advertises exactly that workflow:

```jsonc
"SpriteAnimation": {
  "frames": ["pipe_0001.png", "pipe_0002.png", ...],   // []const []const u8 — fails
  "fps": 6,
  "mode": "loop"
}
"SpriteByField": {
  "component": "TendableWorkstation",
  "field": "level",
  "entries": [                                          // []const Entry — fails
    { "key": 0, "sprite_name": null },
    ...
  ]
}
```

Both silently fail to deserialize, the components don't attach, the ticks skip those entities via their view filter, and there's no log or error — the sprite just doesn't animate. Discovered while writing the end-to-end walkthrough test at `test/example_prefab_animation_walkthrough_test.zig`; that test currently works around the gap by programmatically attaching these components after `spawnFromPrefab`.

## Related components affected today

- `SpriteAnimation.frames: []const []const u8` — array of sprite-name strings.
- `SpriteByField.entries: []const Entry` — each `Entry` is `{ key: i32, sprite_name: ?[]const u8 }`, so this also needs optional handling for the inner string.

## What's needed

1. **Slice type dispatch** in `deserialize(T, value, allocator)`: when `@typeInfo(T)` is `.pointer` with `size == .slice`, get the JSONC value as an array and deserialize each element through the element type, allocating the backing slice into an arena with matching lifetime.
2. **Optional type dispatch**: when `@typeInfo(T)` is `.optional`, `null` JSONC → component field stays `null`; otherwise unwrap and deserialize.
3. **Lifetime / ownership**: the existing `intern_arena` (file-scope, never freed, dedupes `[]const u8`) is the right place — matches how bare strings already survive the parse arena being freed. Slice backings (not string content) need to land there too.

## Out of scope for this ticket

- Deserializing `anyopaque` or pointers-to-single-items — slices and strings only.
- Deep nested slices beyond one level (\`[]const []const []const u8\`); drop in when a use case arrives.

## Follow-ups once the bridge supports this

- Drop the manual re-attach steps in `test/example_prefab_animation_walkthrough_test.zig` and move SpriteAnimation / SpriteByField into `plant.jsonc`.
- Update `rfc/PREFAB-MIGRATION.md` (step 3 of the runtime-overlay migration) so downstream authors can follow the guide exactly without hitting the silent failure.

## Reproducer

Smallest repro is the walkthrough's initial version (before the workaround) — see `test/example_prefab_animation_walkthrough_test.zig` in the \`docs/prefab-animation-walkthrough-example\` branch / PR once that lands.

apotema:
## Summary

Adds a single-file narrative walkthrough of the prefab-foundations + animation chain that landed on `main` today (#474, #482, #483, #484, #485, #475, #476, #480, #481), and along the way unblocks the RFC-advertised workflow by extending the JSONC bridge deserializer to handle optional and slice types.

Closes #487.

### `test/example_prefab_animation_walkthrough_test.zig` (new)

Narrative, single-file, runs under `zig build test`. Reads top-to-bottom as documentation with assertions:

1. Declare a prefab jsonc with `PlantLevel` (saveable driver) + `Sprite` + `SpriteAnimation` + `SpriteByField` inline.
2. `game.spawnFromPrefab("plant", pos)` — root gets `PrefabInstance`, children get `PrefabChild` with `local_path` keyed `children[0]` / `children[1]`.
3. Tick `spriteAnimationTick` + `spriteByFieldTick`. Verify frame advances and field-driven sprite swaps.
4. Save, `resetEcsBackend`, load. Phase 1 respawns the prefab + maps child IDs via `(root, local_path)`; Phase 2 reapplies saved component values; `.transient` animation components come back via the respawn rather than the save file.
5. Verify saveable driver (`PlantLevel`) round-trips at its last-written value, and a single post-load tick re-resolves `SpriteByField` against it.

### `src/jsonc_scene_bridge.zig` — slice + optional deserialization (closes #487)

The bridge's `deserialize` function previously only handled primitives, `[]const u8`, enums, tagged unions, and structs — so slice-of-string and slice-of-struct silently failed. That gated the RFC's entire "declare `SpriteAnimation` inline in a prefab" pitch.

Added two branches to `deserialize`:

- **Optionals** (`?T`): JSONC `null` → optional stays `null` (success, returned as `@as(T, null)`); anything else unwraps and recurses. Supports `?[]const u8` fields like `SpriteByField.Entry.sprite_name`.
- **Slices of non-string types** (`[]const U` where `U != u8`): allocate backing buffer in the existing `intern_arena` (program lifetime, matches the `[]const u8` pattern), deserialize each element recursively. Supports `SpriteAnimation.frames: []const []const u8` and `SpriteByField.entries: []const Entry`.

First-pass bug caught by the walkthrough: the optional handler initially used bare `return null`, which reads as "deserialize failed" on the function's `?T` return. Fixed to `return @as(T, null)` so the caller sees "successfully deserialized a present-but-null optional."

### Drive-by save-policy correction

`SpriteAnimation.save` and `SpriteByField.save` were both declared `.saveable` but their fields (`frames: []const []const u8`, `entries: []const Entry`) aren't serde-writable — the first game to register either component in its `ComponentRegistry` would hit a `@compileError` from the save mixin. Corrected to `.transient`, which matches the RFC's actual design: these come back via Phase 1 prefab respawn, not serde round-trip. Two save-policy tests updated to match.

## Test plan

- [x] `zig build test` green (264/264)
- [x] Walkthrough asserts each step of the pipeline with no manual component attachment — everything flows from the prefab jsonc
- [x] Existing per-PR focused tests unaffected (`spawn_from_prefab_test.zig`, `save_load_two_phase_test.zig`, `jsonc_bridge_prefab_tags_test.zig`, `sprite_animation_tick_test.zig`, `sprite_by_field_tick_test.zig`)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -54,6 +54,7 @@ pub fn build(b: *std.Build) void {
         "test/spawn_from_prefab_test.zig",
         "test/jsonc_bridge_prefab_tags_test.zig",
         "test/save_load_two_phase_test.zig",
+        "test/example_prefab_animation_walkthrough_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/jsonc_scene_bridge.zig
+++ b/src/jsonc_scene_bridge.zig
@@ -1056,26 +1056,25 @@ pub fn JsoncSceneBridge(comptime GameType: type, comptime Components: type) type
             }
 
             // Slices of other types (`[]const Struct`, `[]const []const u8`,
-            // etc.). The `[]const u8` case is handled specifically above
-            // so string deduplication still runs; this branch covers
-            // everything else. Backing storage lands in `intern_arena`
-            // because (a) scene/prefab parse arenas get freed after
-            // load but components borrow these slices for the entity's
-            // lifetime, and (b) we don't want every prefab spawn to
-            // re-allocate identical slices — the shared arena matches
-            // the contract every other cross-parse string already
-            // uses. Each element deserializes recursively, so nested
-            // slices and slice-of-struct both work.
+            // etc.). The `[]const u8` case is handled specifically
+            // above so string deduplication still runs via the
+            // intern pool; this branch covers everything else.
+            //
+            // Lifetime: the caller passes its per-world arena
+            // allocator (see `applyComponent`), which matches the
+            // spawned entity's lifetime and resets on
+            // `resetEcsBackend`. Prior revisions used the file-scope
+            // `intern_arena`, which is never freed — slice contents
+            // aren't deduplicable (unlike bare strings), so every
+            // prefab spawn leaked a new allocation over the process
+            // lifetime. Using the scene-scoped arena instead caps
+            // the cost at "one allocation per entity per scene load,"
+            // released on scene change.
             if (info == .pointer and info.pointer.size == .slice) {
                 const arr = value.asArray() orelse return null;
                 const Element = info.pointer.child;
 
-                if (intern_arena == null) {
-                    intern_arena = std.heap.ArenaAllocator.init(std.heap.page_allocator);
-                    intern_map = std.StringHashMap(void).init(intern_arena.?.allocator());
-                }
-                const arena_alloc = intern_arena.?.allocator();
-                const buf = arena_alloc.alloc(Element, arr.items.len) catch return null;
+                const buf = allocator.alloc(Element, arr.items.len) catch return null;
                 for (arr.items, 0..) |item, i| {
                     buf[i] = deserialize(Element, item, allocator) orelse return null;
                 }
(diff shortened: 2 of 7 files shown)


apotema:
@copilot review

apotema:
@copilot review — addressed gemini L1083: slice backings now go into `active_world.nested_entity_arena` (scene-scoped, freed on `resetEcsBackend`) instead of the program-lifetime `intern_arena`. No more per-spawn leak; allocation is now bounded by "one per entity per scene load," released atomically on scene change. `applyComponent` routes Sprite/Shape/registry-dispatched components through the arena; the transient `stripEntityArrayFields` scratch stays on `game.allocator` since its lifetime is per-call with a matching `defer`.

PR #489, opened 2026-04-23
fix(jsonc-bridge): optional deserialize swallowed inner failures as null

apotema:
## Summary

Follow-up to [#488](https://github.com/labelle-toolkit/labelle-engine/pull/488). Cursor Bugbot flagged a subtle bug in the optional-type branch of `deserialize` at commit 2311b2d that merged anyway — the fix lands here with a regression test.

## The bug

When deserializing a field of type `?U`, the function's return type is `?T` = `??U`. The previous implementation was:

```zig
if (info == .optional) {
    if (value == .null_value) return @as(T, null);
    return deserialize(info.optional.child, value, allocator);
}
```

The recursive call returns `?U`. `return`-ing it from a `??U` function triggers Zig's `?U → ??U` coercion, which wraps the whole `?U` as the outer Optional's payload. That makes two distinct states indistinguishable:

| Inner call returns | Outer coerces to | `deserializeStruct` reads as |
|---|---|---|
| `?U = some(U)` (success) | `??U = some(some(U))` | success, value present ✓ |
| `?U = null` (failure) | `??U = some(null)` | **success, value is null** ✗ |

A prefab declaring `"label": 42` on an `?[]const u8` field (accidental integer, typo'd variant name, etc.) would silently produce a `null` label — indistinguishable from an explicit JSONC `null`.

## Fix

Unwrap the inner result explicitly with `orelse return null` (propagates the outer failure), then re-wrap through `@as(T, inner)` on the success path:

```zig
if (info == .optional) {
    if (value == .null_value) return @as(T, null);
    const inner = deserialize(info.optional.child, value, allocator) orelse return null;
    return @as(T, inner);
}
```

Expanded the docstring around the branch to enumerate the three distinct `null` states so the next author can't make the same mistake without reading the comment.

## Regression test

New `test/jsonc_bridge_deserialize_test.zig` covering:

1. JSONC `null` on an optional field → field is null (success).
2. JSONC string on an optional string field → field is that string (success).
3. JSONC integer on an optional string field → field stays at its default (`null`), and the other fields in the struct still apply.

## Test plan

- [x] `zig build test` green (267/267, +3 new)
- [x] Existing walkthrough test (`example_prefab_animation_walkthrough_test.zig`) still passes — it was the original consumer of the optional deserialize path, via `SpriteByField.Entry.sprite_name: ?[]const u8`.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -55,6 +55,7 @@ pub fn build(b: *std.Build) void {
         "test/jsonc_bridge_prefab_tags_test.zig",
         "test/save_load_two_phase_test.zig",
         "test/example_prefab_animation_walkthrough_test.zig",
+        "test/jsonc_bridge_deserialize_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/jsonc_scene_bridge.zig
+++ b/src/jsonc_scene_bridge.zig
@@ -1029,16 +1029,35 @@ pub fn JsoncSceneBridge(comptime GameType: type, comptime Components: type) type
             // instead of getting routed into the string branch with a
             // null value and failing.
             //
-            // Note the two distinct `null`s at play: the function
-            // returns `?T` to signal "deserialize failed," but for an
-            // optional field type the *success* case of "JSON value
-            // was null" is a valid `T` — so we return `@as(T, null)`,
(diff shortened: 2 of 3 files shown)


apotema:
@copilot review

PR #491, opened 2026-04-24
feat(save_load): auto-collect PrefabInstance/PrefabChild entities

apotema:
## Summary

Closes a gap I flagged in \`Game.spawnFromPrefab\`'s docstring during #488 and surfaced for real when flying-platform-labelle adopted the prefab-foundations arc (\`flying-platform-labelle#291\`).

## The gap

\`saveGameState\` collected entities by iterating registered components with \`.saveable\` / \`.marker\` save policies. A prefab whose root carried ONLY a \`Sprite\` + the engine-auto-attached \`PrefabInstance\` (common shape — \"pure visual\" prefabs like a scene background) was invisible to that pass — missed from the save file, no Phase 1 respawn on load, entity gone. Before this fix, game authors had to sprinkle a marker component solely to placate save/load, OR manually re-create the entity in their F9 handler (what FP does for \`background_sky\` today).

## The fix

\`saveGameState\` now runs a second sweep after the registry-driven pass, collecting entities that carry \`PrefabInstance\` or \`PrefabChild\` and adding them to the same dedup set. Both funnel through the existing \`entity_set\` so duplicates are squashed.

Why both tag types:

- **\`PrefabInstance\`** — main case. Every prefab root now survives save/load regardless of other components.
- **\`PrefabChild\`** — usually redundant (prefab descendants typically carry game-owned saveables that already got them collected), but catches the edge case where a prefab-declared child has no game state AND is referenced by another saved entity via its entity ID. Phase 1b's \`(root, local_path)\` remap can't populate \`id_map\` without the child's saved entry.

## Regression test

\`test \"save: entities with only PrefabInstance are collected (no game-owned saveable)\"\` in \`save_load_two_phase_test.zig\`:

1. Prefab with only \`Position\` (no registered component).
2. \`spawnFromPrefab\` → entity gets \`PrefabInstance\` auto-attached but nothing else that would have triggered collection pre-fix.
3. \`saveGameState\` → assert the save file mentions the prefab path (would have been empty before this change).
4. \`resetEcsBackend\` + \`loadGameState\` → exactly 1 entity back with \`PrefabInstance.path == \"pure_visual\"\`.

## Docstring

Updated \`Game.spawnFromPrefab\`'s docstring — the \"authors need at least one saveable or marker component\" caveat is gone. Replaced with a note about the collection guarantee and a more accurate caveat about Phase 2 override precedence.

## Downstream

flying-platform-labelle can drop the manual \`background_sky\` re-creation in its F9 handler. That's a game-side commit gated on this engine release shipping.

## Test plan

- [x] \`zig build test\` green (268/268, +1 new)
- [x] Existing prefab / save-load tests unchanged — no collection-path regressions

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -540,18 +540,24 @@ pub fn GameConfig(
         /// structured-overrides pipeline lands in a follow-up and will
         /// dupe `overrides` into the same arena then.
         ///
-        /// **Interaction with the save mixin:** `spawnFromPrefab` tags
-        /// the entity so Phase 1 on load can respawn it from its
-        /// prefab. But the save mixin only *collects* entities that
-        /// carry at least one component with `.saveable` or `.marker`
-        /// save policy (see `saveGameState` in
-        /// `src/game/save_load_mixin.zig`). A prefab-spawned entity
-        /// that carries only engine built-ins (PrefabInstance/
-        /// PrefabChild) + non-saveable components (e.g. Sprite) will
-        /// NOT appear in the save file and therefore won't be respawned
-        /// on load. Authors relying on save/load round-trip need at
-        /// least one saveable or marker component on the prefab root —
-        /// typically game-owned (e.g. `Workstation`, `RoomDecor`).
+        /// **Interaction with the save mixin:** tagged entities
+        /// always survive `saveGameState` → `loadGameState` round-
+        /// trip. Since the sweep in `saveGameState` auto-collects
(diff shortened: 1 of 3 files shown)


apotema:
@copilot review

apotema:
@copilot review — addressed gemini L289: the two prefab-tag sweep blocks collapsed into a single `inline for (.{ Game.PrefabInstanceComp, Game.PrefabChildComp }) |Tag|`. Same comptime unrolling, half the source lines.

PR #492, opened 2026-04-24
feat(hooks): scene_before_reset — last-chance hook before ECS wipe

apotema:
## Summary

Adds a new engine hook event that fires right before \`resetEcsBackend\` on the two atomic scene-tear-down paths (\`setSceneAtomic\` and \`loadGameState\`). Plugin controllers with per-world heap state — FP uses this pattern in 8 places (\`pathfinder\`, \`scheduler\`, \`caretaker\`, \`needs_machine\`, \`command_buffer\`, \`behavior_tree\`, \`worker_controller\`, \`production\`) — can listen and free their state before the ECS is wiped, instead of leaking allocations across every F9 load or panicking on F8 scene restart.

## Why

Addresses [flying-platform-labelle#290](https://github.com/Flying-Platform/flying-platform-labelle/issues/290). Plugin controllers anchor a heap-allocated \`State\` struct via a \`state_ptr: usize\` field on a singleton ECS component. When the ECS is reset:

- The singleton entity is destroyed → \`state_ptr\` component vanishes.
- The heap \`State\` is **orphaned** — nothing freed it.
- \`findState(game)\` returns \`null\` for every subsequent call.

FP has had a manual F9 bracket in \`save_load.zig\` that calls each plugin's \`Controller.deinit\` before \`loadGameState\` and \`Controller.setup\` after — but only for 2 of the 8 plugins. The other 6 leak silently on every F9. And F8 (scene restart) can't apply the bracket game-side because \`setSceneAtomic\` runs deferred inside \`Game.tick\`'s \`pending_scene_change\` processing — game scripts can't interpose.

## Shape

New variant \`scene_before_reset: SceneInfo\`. Emit sites:

| Path | Hook fires at | Name payload |
|---|---|---|
| \`setSceneAtomic(name)\` | Just before \`resetEcsBackend\` | **Outgoing** scene (about to be torn down) |
| \`loadGameState(path)\` | Just before \`resetEcsBackend\` | Current scene (same in/out — world reloads from save file, scene doesn't change) |

Not emitted from non-atomic \`setScene\` — that path destroys entities individually through \`unloadCurrentScene\`, and plugins handling per-entity teardown should listen to \`entity_destroyed\` rather than \`scene_before_reset\`. Pinning that boundary with a dedicated test keeps it from drifting.

## Contract for listeners

- Free per-world heap state in \`scene_before_reset\` (while the singleton still exists so \`findState\` can locate the ptr).
- Re-allocate fresh state in the existing \`scene_load\` handler after the new scene loads.

## Downstream follow-up

FP can drop the manual bracket in \`save_load.zig\` and replace it with a single \`hooks/plugin_controllers_hooks.zig\` that covers all 8 plugins. Fixes both #290's F8 panic and the silent leak of 6 plugins' state on every F9 cycle.

## Test plan

- [x] \`zig build test\` green — 271/271 (+4 new across both review revisions)
- [x] \`scene_before_reset: fires on setSceneAtomic before the ECS wipe\` — hook is emitted once per atomic swap and strictly precedes \`scene_before_load\`
- [x] \`scene_before_reset: does NOT fire on non-atomic setScene\` — pins the atomic-only boundary

### DIFF
--- a/src/game/save_load_mixin.zig
+++ b/src/game/save_load_mixin.zig
@@ -470,6 +470,18 @@ pub fn Mixin(comptime Game: type) type {
             //     and hit `destroyEntityOnly` with invalid handles.
             self.scene_entities.clearRetainingCapacity();
             self.clearActiveSceneEntities();
+            // `scene_before_reset` fires for the same reason
+            // `setSceneAtomic` emits it: plugin controllers with
+            // per-world heap state must free it before
+            // `resetEcsBackend` destroys the singleton component
+            // that holds their pointer. Listeners that also handle
+            // the F8 path see a consistent bracket across both
+            // reset entry points. The scene name here is the
+            // scene that's about to be reloaded (same name in,
+            // same name out — the ECS is wiped and reloaded from
+            // the save file, not swapped to a different scene).
+            const current_scene = self.current_scene_name orelse "";
+            self.emitHook(.{ .scene_before_reset = .{ .name = current_scene } });
             self.resetEcsBackend();
 
             // Step 2: Create new entities and build ID map.
--- a/src/game/save_load_mixin.zig
+++ b/src/game/save_load_mixin.zig
@@ -468,8 +468,6 @@ pub fn Mixin(comptime Game: type) type {
             //     `spawnFromPrefab` then appends new IDs on top, and a
             //     later `unloadCurrentScene` would iterate the stale ones
             //     and hit `destroyEntityOnly` with invalid handles.
-            self.scene_entities.clearRetainingCapacity();
-            self.clearActiveSceneEntities();
             // `scene_before_reset` fires for the same reason
             // `setSceneAtomic` emits it: plugin controllers with
             // per-world heap state must free it before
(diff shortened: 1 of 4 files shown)


apotema:
@copilot review

apotema:
@copilot review — addressed:

- **Gemini HIGH + Bugbot MEDIUM** (same root cause): hook was firing after `unloadCurrentScene` had already destroyed tracked entities individually. Moved the emit BEFORE any entity destruction and aligned the ordering with `loadGameState` — clear tracking lists, emit hook, `resetEcsBackend` wipes atomically. Listeners now see the full pre-teardown world on both paths.
- **Copilot L367 + L484** (null-name payload): added `if (name) |n|` guards at both emit sites. No more empty-string sentinel for listeners to handle.
- **Copilot L268** (test assertion strength): tightened from `saw_before_reset = true` bool to explicit count == 1. Added a companion test guarding the no-emit-on-first-swap contract.

271/271 tests green (+1).

apotema:
@gemini review

apotema:
@copilot review — addressed:

- **Gemini HIGH** (OOM-robust cleanup): switched the emit source from the `previous_name` dupe to `self.current_scene_name` directly. The dupe's `catch null` would silently skip the cleanup hook under memory pressure — exactly when listeners need to run. `current_scene_name` survives OOM on other allocations.
- **Copilot L325** (test comment hygiene): rewrote the stale "Copilot L367 guard" comment to describe the actual contract without tool/line references.
- **Copilot L229** (test count in PR description): updated the PR body; the delta is +3 new tests on this PR, not +2 as originally said.

PR #300, opened 2026-02-20
fix: migrate deprecated .backend to .targets in project.labelle

apotema:
## Summary
- Replace deprecated `.backend = .X` with `.targets = .{.X_desktop}` in 5 project.labelle files
- Remove redundant `.ecs_backend = .zig_ecs` (default) from example_entity_refs and example_parent_refs
- Affected examples: example_2, example_bgfx, example_entity_refs, example_parent_refs, example_sokol

## Test plan
- [ ] `CI_TEST=1 zig build run` in each affected example directory
- [ ] CI passes

Closes #293

PR #494, opened 2026-04-27
feat(gizmos): honor setGizmosEnabled for declarative *.zon gizmos

### LINKED ISSUE #493: Gate gizmos/*.zon Shape entities behind setGizmosEnabled
## Problem

The \`gizmos/*.zon\` files in a labelle game project are presented as
\"gizmos\" — but they don't actually flow through the gizmo toggle.
\`JsoncSceneBridgeWithGizmos.reconcileGizmos\` instantiates them as
persistent ECS entities with \`Shape\` components, which render
through the regular sprite/shape pipeline. \`renderGizmos\` (in
\`game/gizmo_mixin.zig:117\`) only gates the imperative
\`drawGizmo*Category\` draw-queue, so flipping \`setGizmosEnabled\`
has no effect on these entities.

This bit a flying-platform-labelle player today — EIS/IIS/IOS/EOS
outline rectangles and a workstation triangle were drawing on top of
room art in normal play, with no way to toggle them off. The repo's
workaround was deleting the offending files (flying-platform-labelle PR #300)
because there's no way to keep them around as a debug aid.

## Why this matters

The naming (\`gizmos/\` directory, \"gizmo\" in everything that handles
them) primes any user to expect \`G\` to toggle them. The gap is a footgun:

- Authors ship a \`.zon\` debug overlay, expect the toggle to work,
  end up with permanent visual noise.
- \`drawGizmo*\`-driven scripts already gate correctly. The \`.zon\`
  versions silently don't.
- Forces a binary choice today: always on, or delete entirely.

## Proposed fix

\`reconcileGizmos\` should listen for the gizmo toggle and either:

1. **Per-Shape visibility flag** — flip a flag on the shape entity in
   response to \`setGizmosEnabled\`. Cheap to implement, keeps
   entities alive across toggles.

2. **Spawn/despawn on toggle** — \`reconcileGizmos\` creates entities
   only when \`gizmos_enabled\` is true; tears them down on the off
   transition. More churn, cleaner lifetime.

Either way the contract becomes: \`gizmos/*.zon\` is a debug overlay
gated by \`G\`, same as \`drawGizmo*\` calls.

## Acceptance

- \`setGizmosEnabled(false)\` hides every entity created from a
  \`gizmos/*.zon\` file.
- \`setGizmosEnabled(true)\` brings them back identical to the
  always-on state today.
- No regressions for existing imperative \`drawGizmo*\` callers.
- Doc note that declarative gizmos respect the same toggle as the
  imperative ones.

## Reference

- flying-platform-labelle PR #300 — user-visible symptom.
- \`src/game/gizmo_mixin.zig:117\` — current toggle.
- \`src/jsonc_scene_bridge.zig:1316\` — \`reconcileGizmos\` site.

apotema:
Closes #493.

## Problem

Declarative gizmos defined via `gizmos/*.zon` spawn ECS entities with `Shape` / `Sprite` components and render through the normal sprite/shape pipeline. `renderGizmos` only gates the imperative `drawGizmo*` queue, so `setGizmosEnabled(false)` had no effect on `.zon`-defined gizmos — they always rendered. Authors of these files reasonably expect `G` to toggle them; the gap was a footgun (flying-platform-labelle hit it and worked around with deletion in PR #300).

## Fix

`reconcileGizmos` now runs a small `syncGizmoVisibility` sweep after the per-gizmo create pass. Every entity carrying the internal `Gizmo` component has its `Shape.visible` / `Sprite.visible` flipped to match `gizmos_enabled`. Entities aren't destroyed — keeping them across toggles avoids churn, save/load surprises, and respects the user's clarification "we may need to show them later."

Steady-state cost: one ECS view scan per frame (already happens for the create pass) plus a per-entity visibility check. The `markVisualDirty` call is gated behind a `visible != target` test, so a frame with nothing to change is a no-op for the renderer.

## Test plan

- [x] `zig build test` — full engine test suite green (no regressions).
- [ ] End-to-end smoke in flying-platform-labelle:
  1. Restore the deleted `gizmos/{eis,iis,ios,eos,workstation}.zon` files (#300 workaround).
  2. Bump the engine pin to this PR's hash.
  3. Run `labelle run`. Gizmos start visible (default `gizmos_enabled = true`).
  4. Press `G` — all five `.zon` shapes hide together.
  5. Press `G` again — they come back, identical to step 3.

No new unit test in this PR: `JsoncSceneBridgeWithGizmos` has no existing fixture and bootstrapping one for a single-line behaviour check would be disproportionate. Open to adding it if the gizmo path picks up further behavior we want pinned.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -57,6 +57,7 @@ pub fn build(b: *std.Build) void {
         "test/example_prefab_animation_walkthrough_test.zig",
         "test/jsonc/bridge_deserialize_test.zig",
         "test/jsonc/deserializer_test.zig",
+        "test/jsonc_bridge_gizmo_visibility_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/jsonc_scene_bridge.zig
+++ b/src/jsonc_scene_bridge.zig
@@ -50,6 +50,15 @@ pub fn JsoncSceneBridgeWithGizmos(
     const Gizmo = core.GizmoComponent(Entity);
     const Writer = scene_mod.EntityWriter(GameType, Components);
     const GizmoAttached = struct { _: u8 = 1 };
+    // Tags a gizmo entity whose `visible` flag was force-flipped to
+    // `false` by `syncGizmoVisibility` because the global toggle is
+    // off. On toggle-on, only entities carrying this marker get
+    // their visibility restored — author-set
+    // `visible = false` (declared in a `gizmos/*.zon`) stays
+    // hidden across toggles, matching the principle of least
+    // surprise that an explicit declaration outranks the global
+    // override.
+    const GizmoForcedHidden = struct { _: u8 = 1 };
     const BaseBridge = JsoncSceneBridge(GameType, Components);
 
     return struct {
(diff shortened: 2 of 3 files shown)


apotema:
@gemini review

apotema:
@gemini review

PR #525, opened 2026-05-13
feat(assets): shared loader payload types for Phase 4 (#447, #448)

apotema:
## Summary

- Fills in the `audio:` + `font:` variants of `DecodedPayload` / `UploadedResource` in `src/assets/loader.zig` so the audio loader ([#447](https://github.com/labelle-toolkit/labelle-engine/issues/447)) and font loader ([#448](https://github.com/labelle-toolkit/labelle-engine/issues/448)) implementation PRs can land in parallel without colliding on this file. No loader logic changes — the stubs in `loaders/audio.zig` and `loaders/font.zig` still return `error.NotImplemented`; this PR is type plumbing only.
- Adds `src/font_types.zig` with `FontId` / `Glyph` / `CodepointEntry` / `KernPair`, mirroring the `audio_types.zig` shape. Re-exported via `root.zig`. Exercised by `test/font_types_test.zig` (size-stability locks + handle-validity round-trip).
- Adds `WorkRequest.params` and `AssetEntry.params` as type-erased `?*const anyopaque` so loaders that need decode-time parameters can ride the existing worker pipeline. Image and audio register-sites pass `null`; fonts wire this up in the font-loader PR per [RFC-FONT-LOADER §7](https://github.com/labelle-toolkit/labelle-engine/pull/524).
- Module plumbing: `audio_types` and `font_types` promoted to named modules so the same import resolves identically from both the engine module root (`src/root.zig`) and the standalone assets-tests root (`src/assets/mod.zig`).

Unblocks parallel work on [#463](https://github.com/labelle-toolkit/labelle-engine/pull/463) (audio loader RFC) and [#524](https://github.com/labelle-toolkit/labelle-engine/pull/524) (font loader RFC) implementations. Should merge before either implementation PR.

## Test plan

- [x] `zig build test` — 350/350 passing (3 new in `font_types_test.zig`).
- [ ] Confirm the `WorkRequest.params` lifetime contract reads right (borrowed at register time, lives as long as the entry).
- [ ] Sanity-check that the named-module promotion of `audio_types` / `font_types` doesn't break any downstream consumer that imports them by path (`src/audio.zig` and `src/root.zig` updated; no other path imports found).

PR #524, opened 2026-05-13
docs: RFC for font loader — Phase 4 of asset streaming (#448)

apotema:
## Summary

- Drafts `RFC-FONT-LOADER.md`, the sibling Phase 4 RFC to [#463](https://github.com/labelle-toolkit/labelle-engine/pull/463) (audio loader, #447). Replaces the current `src/assets/loaders/font.zig` stub with an `stb_truetype`-based loader: 8-bit alpha glyph atlas baked off-thread, GPU upload + `FontId` on main thread, refcount-driven unload through the existing `AssetCatalog`.
- Mirrors the audio RFC's structure (problem / goals / non-goals / numbered design sections / migration plan / open questions / references) and borrows its conventions verbatim where they apply (inline `DecodedPayload` variants, backend-hook function-pointer struct, `clearBackend` test seam, `asset_failure_policy` integration).
- Identifies the three places fonts diverge from the audio shape: bake params (`pixel_height`, codepoint ranges, atlas size) carried through a new `WorkRequest.params` field; a `registerFont` entry point so image / audio call sites stay untouched; and a minimum GUI hook-up step in the migration plan since fonts have no existing engine surface today (unlike audio's `SoundId` / `AudioInterface`).

Closes #448 once Phase 4 implementation lands.

## Test plan

- [ ] Review structural parity with `RFC-AUDIO-LOADER.md` from #463 — inline payloads, backend hook shape, vtable body, sentinel-test recipe, references list.
- [ ] Confirm the 7 open questions cover everything reviewers would need to decide before implementation starts (bitmap vs SDF, default glyph range, cache key shape, `WorkRequest.params` typing, GUI consumption site, variable-font scope, high-DPI baking).
- [ ] Sanity-check the citations against the current working tree (`src/assets/loader.zig`, `src/assets/loaders/font.zig`, `src/gui_types.zig`, `src/game/gui_mixin.zig:52`).

Issue #448, opened 2026-04-15
tracking: font loader — Phase 4 of asset streaming


apotema:
Tracking issue for Phase 4 of the [Asset Streaming RFC](https://github.com/labelle-toolkit/labelle-engine/pull/437).

Real `decodeFont` + glyph rasterise replaces the stub loader landed in Phase 1. Separate RFC required per the RFC Migration §Phase 4.

## Out of scope for this ticket
Streaming machinery is already done by the time Phase 4 starts. This ticket only needs:
- TTF/OTF decoder (stb_truetype likely)
- Glyph atlas rasterise on `upload` / `free`
- Manifest support for font names (already validated by the assembler asset-name validation ticket)

## Next step
Draft follow-up RFC before implementation.

apotema:
## Phase 4 font loader — status rollup

Engine-side work for the font loader (Phase 4 of the asset streaming roadmap) has landed on `main`. This tracking issue stays open until the cross-repo follow-ups complete.

### Merged in labelle-toolkit/labelle-engine
- #524 — `docs(rfc): font loader — Phase 4 of asset streaming`
- #525 — `feat(assets): shared loader payload types for Phase 4` (audio + font)
- #529 — `feat(assets+gui): real font loader vtable + Label.font hookup — Phase 4`

### In flight / open
- labelle-toolkit/labelle-gfx#258 — `feat(backend): font decode/upload/unload traits — Phase 4` (awaiting Copilot review)
- labelle-toolkit/labelle-assembler — `feat(codegen): writeFontBackendWiring scaffolding — Phase 4` (PR being opened in parallel; in flight, PR number pending)

### Cross-repo follow-ups still outstanding
- Concrete raylib font backend adapter against the gfx font traits (once #258 lands)
- Concrete sokol font backend adapter against the gfx font traits
- Assembler call sites that actually invoke `writeFontBackendWiring` (once the codegen PR merges)

Closing this tracking issue once the concrete backend adapters and the assembler call sites are wired up end-to-end.

apotema:
Phase 4 font chain is complete end-to-end. Final state:

**labelle-engine (merged):**
- #524 — font loader RFC
- #525 — shared loader payload types
- #529 — font loader vtable + Label.font GUI hookup
- #532 — Game.registerFont/loadFont/loadFontIfNeeded helpers
- #533 (still open) — extern struct for Glyph/CodepointEntry/KernPair

**labelle-gfx (merged):**
[...]

PR #527, opened 2026-05-13
feat(assets+gui): real font loader vtable + Label.font hookup — Phase 4 (#448)

apotema:
## Summary

Phase 4 of the asset-streaming RFC for the **font loader**. Replaces the
font-loader stub with a working decode/upload/drop/free vtable, threads
`FontBakeParams` from `register` through the worker to the loader, and
gives `Label` an optional `font: ?[]const u8` field that resolves to a
baked `FontId` at render time.

Links:
- RFC PR: #524 (RFC-FONT-LOADER.md)
- Shared payload types PR: #525 (parent branch — PR retargets to `main`
  after #525 merges)
- Ticket: #448

## What changed

* `src/assets/loaders/font.zig`: real vtable + `FontBackend`,
  `FontBakeParams`, `CodepointRange`, `DecodedFont`, six in-file mock
  backend tests covering decode/upload/drop/free + error paths.
* `src/assets/catalog.zig`: `registerFont(name, file_type, bytes, params)`
  + `registerWithParams` helpers. The original `register` is now a thin
  wrapper that forwards `params = null`.
* `src/gui_types.zig`: `Label.font: ?[]const u8 = null`.
* `src/game/gui_mixin.zig`: thread `*Game` through view rendering; when
  `lbl.font` is set, look the asset up via
  `game.assets.entries.getPtr(...).resource.?.font` and dispatch to
  `Gui.labelWidgetWithFont`. Falls back to today's default-font
  `Gui.labelWidget` path when `font == null`.
* `src/root.zig` + `src/assets/mod.zig`: re-export `FontLoader`,
  `FontBackend`, `FontBakeParams`, `CodepointRange`, `DecodedFont`.
* `test/font_loader_test.zig` + `build.zig`: round-trip catalog test
  against a mock `FontBackend` and a second test that registers two
  fonts under different `pixel_height` and verifies the params pointers
  stay independent through the worker round-trip.

## RFC divergence — vtable.decode signature

The RFC / task description says the worker forwards `WorkRequest.params`
to `vtable.decode`, but on the parent branch the vtable's `decode`
signature is `(file_type, data, allocator) → DecodedPayload` — no
params arg, so the worker had no way to pass it through. To make the
font loader reachable, this PR extends the signature to
`(file_type, data, params: ?*const anyopaque, allocator) → DecodedPayload`.
The image and audio decode bodies just `_ = params;` it; the font
decode `@ptrCast`s back to `*const FontBakeParams`.

This is a documented breach of the "do not modify `loader.zig` /
`worker.zig`" constraint in the task brief — flagged here so the
reviewer can confirm the parent payload-types PR (#525) should
absorb this signature change before the merge dance, or whether to
keep it on this PR.

## Catalog changes

Yes — added `registerFont` and `registerWithParams` to
`src/assets/catalog.zig`. `register` itself is now a one-line forward
to `registerWithParams(.., null)`, preserving every existing call site
unchanged.

## Test plan

- [x] `zig build test` (full engine suite) — passes
- [x] New `test/font_loader_test.zig`: 2 catalog round-trip tests
- [x] In-file `src/assets/loaders/font.zig`: 6 mock-backend tests
- [x] Existing image-loader + catalog tests still green (signature
      change to `vtable.decode` propagated)
- [x] `single_threaded` (WASM) assets test compiles
- [ ] Manual smoke test from a flying-platform-labelle scene — pending
      a `FontBackend` adapter in labelle-gfx (#449 follow-up)

## Notes

- `Gui.labelWidgetWithFont` is referenced but does not yet exist on
  any backend. Per the task brief's "keep this minimal — the renderer
  doesn't need a full text styling rework" guidance, the call site
  uses the `@hasDecl`-guarded `Gui` wrapper convention so backends
  without the method still link cleanly. The labelle-gfx adapter will
  add the concrete implementation alongside the `FontBackend` baker.
- All four decoded slices (`bitmap`, `glyphs`, `codepoint_index`,
  `kerning`) follow the same allocator-ownership contract as the
  image loader's `pixels`: `upload` frees on success, `drop` frees
  on the discard path, `testing.allocator` catches any leak.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -61,6 +61,7 @@ pub fn build(b: *std.Build) void {
         "test/scene_ref_test.zig",
         "test/asset_catalog_test.zig",
         "test/font_types_test.zig",
+        "test/font_loader_test.zig",
         "test/asset_streaming_shim_test.zig",
         "test/animation_def_test.zig",
         "test/sprite_animation_test.zig",
--- a/src/assets/catalog.zig
+++ b/src/assets/catalog.zig
@@ -196,6 +196,23 @@ pub const AssetCatalog = struct {
         file_type: [:0]const u8,
         bytes: []const u8,
     ) !void {
+        return self.registerWithParams(name, loader_kind, file_type, bytes, null);
+    }
+
+    /// Metadata + decode-parameters registration. Same lifetime
+    /// contract as `register`: every borrowed slice (`name`,
+    /// `file_type`, `bytes`) AND the `params` pointer must outlive the
+    /// catalog entry. Used by `registerFont` to attach a
+    /// `*const FontBakeParams` so the worker can hand it to the font
+    /// loader's `decode` via `WorkRequest.params`.
+    pub fn registerWithParams(
+        self: *AssetCatalog,
+        name: []const u8,
+        loader_kind: LoaderKind,
+        file_type: [:0]const u8,
+        bytes: []const u8,
+        params: ?*const anyopaque,
+    ) !void {
         if (self.entries.contains(name)) return error.AssetAlreadyRegistered;
         try self.entries.put(name, .{
             .state = .registered,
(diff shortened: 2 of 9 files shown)


apotema:
@copilot review

PR #532, opened 2026-05-13
feat(game): registerSound + registerFont helpers — Phase 4 (#447, #448)

apotema:
## Summary

Six new methods on `Game`, mirroring the existing `registerAtlasFromMemory` / `loadAtlasFromMemory` / `loadAtlasIfNeeded` trio so the assembler can emit consistent register/load calls for `.wav`/`.ogg`/`.ttf`/`.otf` resources alongside today's `.png` atlases:

- `registerSoundFromMemory(name, file_type, audio_data)`
- `loadSoundFromMemory(name, file_type, audio_data)`
- `loadSoundIfNeeded(name)`
- `registerFontFromMemory(name, file_type, font_data, params)`
- `loadFontFromMemory(name, file_type, font_data, params)`
- `loadFontIfNeeded(name)`

Same surface contract as the atlas helpers: `register*` is deferred (catalog stores the bytes; decode + upload run on the first `acquire`); `load*FromMemory` is eager (register + busy-pump); the `IfNeeded` path bumps the catalog refcount with the matching errdefer-release + `resetFailed` pattern.

Font helpers thread a `*const FontBakeParams` pointer through `catalog.registerFont`; the loader casts it back from `?*const anyopaque` inside its `decode` (RFC-FONT-LOADER §7).

Both helpers swallow `error.AssetAlreadyRegistered` to match the atlas pattern — assembler-generated scene manifest code may have registered the same name directly on the catalog before the shim fires.

## Why now

This is the engine half of the third follow-up listed in [labelle-assembler#102][a102] and [labelle-assembler#103][a103]. The next assembler PR will grow `ProjectConfig.resources` to accept audio + font resource shapes and dispatch the codegen between `loadAtlasFromMemory` and these new helpers based on the file extension. Without these helpers, the assembler would have to emit raw `game.assets.register(...)` / `game.assets.registerFont(...)` calls, which works but breaks symmetry with the image path.

[a102]: https://github.com/labelle-toolkit/labelle-assembler/pull/102
[a103]: https://github.com/labelle-toolkit/labelle-assembler/pull/103

## Test plan

- [x] `zig build test` — 373/373 passing (+7 new in `test/asset_streaming_shim_test.zig`).
- [x] New tests cover: register+load round-trip, eager variant, idempotent re-load, double-register tolerance (audio), distinct-params produce distinct entries (font).
- [x] `MockFont.decodeFn` captures `params.pixel_height` so the shim's pointer-threading through `catalog.registerFont` is verified end-to-end.
- [ ] Once merged: labelle-assembler PR for `ProjectConfig` audio/font resource shape + extension dispatch (separate PR).

## References

- #447 — audio loader tracking
- #448 — font loader tracking
- labelle-assembler#102 — audio codegen scaffolding (calls these helpers' Sound counterparts in its emitted adapter)
- labelle-assembler#103 — font codegen scaffolding

### DIFF
--- a/src/game.zig
+++ b/src/game.zig
@@ -1845,12 +1845,23 @@ pub fn GameConfig(
             _ = try self.loadSoundIfNeeded(name);
         }
 
-        /// Decode + upload a previously-registered sound if it hasn't
-        /// already been loaded. No-op (returns `false`) on already-ready
-        /// entries. Mirrors `loadAtlasIfNeeded`'s busy-pump shape; see
-        /// the long-form rationale there for the `errdefer release`,
-        /// `resetFailed`, and `std.Thread.yield` choices.
-        pub fn loadSoundIfNeeded(self: *Self, name: []const u8) !bool {
+        /// Shared busy-pump used by `loadSoundIfNeeded` and
+        /// `loadFontIfNeeded`. The atlas counterpart
+        /// (`loadAtlasIfNeededImpl`) does NOT delegate here because
+        /// it threads a `markPendingLoaded` call into the
+        /// `TextureManager` after the upload — the audio + font paths
+        /// have no equivalent post-upload bridging.
+        ///
+        /// Lifecycle (see `loadAtlasIfNeededImpl` for the long-form
+        /// rationale this mirrors): bumps the catalog refcount so the
+        /// worker enqueues the decode, then busy-pumps until the entry
+        /// is `.ready` (happy path) or `lastError` is set (decode /
+        /// upload failed). `errdefer release` keeps the refcount
+        /// consistent on every failure path so a retry after a failed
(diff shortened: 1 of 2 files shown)


apotema:
Addressed in `7ed07d3` — extracted `loadAssetIfNeededInternal` as a private helper shared by `loadSoundIfNeeded` and `loadFontIfNeeded`. Both methods are now one-liners delegating to it.

Atlas's `loadAtlasIfNeededImpl` deliberately doesn't delegate — it threads `markPendingLoaded` into the `TextureManager` after the upload, which audio/font don't have. Better than reshaping the helper to return optional post-upload work; keeps both call sites readable.

`zig build test`: 373/373 still passing.

Issue #107, opened 2025-12-30
Add Game reference to ComponentPayload for component lifecycle callbacks


apotema:
## Summary

Currently, `ComponentPayload` only provides `entity_id` in component lifecycle callbacks (`onAdd`, `onSet`, `onRemove`). This limits what components can do in their callbacks since they cannot access the registry or other game state.

## Problem

When a component defines an `onAdd` callback, it only receives the entity ID:

```zig
pub fn onAdd(payload: engine.ComponentPayload) void {
    // Only have payload.entity_id
    // Cannot access registry to query component data
    // Cannot access other game state
}
```

This makes it impossible to:
- Access the component's own data (e.g., list nested entity references)
- Query other components on the same or related entities
- Access game-level services or state

## Proposed Solution

Add a `game: *Game` pointer to `ComponentPayload`:

```zig
pub const ComponentPayload = struct {
    entity_id: u64,
    game: *anyopaque,  // Pointer to Game (opaque to avoid circular deps)
};
```

This would allow components to:

```zig
pub fn onAdd(payload: engine.ComponentPayload) void {
    const game = payload.getGame();  // Helper to cast from anyopaque
    const registry = game.getRegistry();
    
    // Now can access component data
    if (registry.tryGet(Workstation, entity)) |ws| {
        for (ws.input_storages) |storage_entity| {
            std.log.info("Input storage: {any}", .{storage_entity});
        }
    }
}
```

## Use Case

In the bakery-game, `Workstation` components have an `input_storages: []const Entity` field. The `onAdd` callback should be able to log these nested entities, but currently cannot access the component data.

## Implementation Notes

- Use `*anyopaque` for the Game pointer to avoid circular import issues
- Provide a helper method like `getGame()` that casts to the correct type
- Update ECS adapters (zig_ecs, zflecs) to pass the Game reference when invoking callbacks

Issue #105, opened 2025-12-30
labelle upgrade fails with HTTP 400 error on some installations


apotema:
## Description

The `labelle upgrade` command fails with HTTP 400 error when trying to download the new binary, despite redirect handling being implemented.

## Error Output

```
Checking for updates...
Current version: 0.26.0
Latest version:  v0.28.0

Downloading labelle v0.28.0 for darwin-arm64...
HTTP error: 400 bad_request
Download failed: error.HttpStatusError
Error downloading binary: error.DownloadFailed
error: DownloadFailed
```

## Investigation

- The v0.26.0 release binary downloaded fresh from GitHub works correctly
- The user's installed binary reports v0.26.0 but has a different MD5 hash than the release
- This suggests the user has a locally-built binary that may have been built from a different commit

## Root Cause Hypothesis

When the user previously ran `labelle upgrade` from v0.25.0, it may have failed and they built locally from source. The locally-built binary may have version 0.26.0 but could be missing the redirect handling fix or have some other difference.

## Workaround

Users experiencing this issue can manually upgrade:

```bash
curl -sL "https://github.com/labelle-toolkit/labelle-engine/releases/download/v0.28.0/labelle-darwin-arm64.tar.gz" | sudo tar xz -C /usr/local/bin
```

(Replace `darwin-arm64` with appropriate platform)

## Suggested Improvements

1. **Add verbose mode**: `labelle upgrade --verbose` to show more details about the HTTP request/response
2. **Better error messages**: Include the URL being fetched when an error occurs
3. **Retry logic**: Implement automatic retry on transient HTTP failures
4. **Checksum verification**: Verify downloaded binary matches expected hash before installation

## Related

This is related to the redirect handling fix added in commit 7e197b9.

PR #533, opened 2026-05-13
refactor(font_types): make POD structs extern for cross-repo bitcast

apotema:
## Summary

\`Glyph\`, \`CodepointEntry\`, and \`KernPair\` are three structurally-identical-but-nominally-distinct types defined in three places:

- \`labelle-engine/src/font_types.zig\` (this PR)
- \`labelle-gfx/src/backend.zig\` (sibling PR coming)
- Concrete backends (the first one being labelle-assembler#108's sokol impl)

The assembler-generated \`FontBackendAdapter\` (from #525 / labelle-assembler#103+#105+#107) marshals between them at the codegen boundary. For flat fields the marshal is a plain \`.advance = d.advance\` assignment. For slice fields (\`.glyphs\`, \`.codepoint_index\`, \`.kerning\`) the fix is a zero-cost \`@ptrCast\` between \`[]BackendGfx.Glyph\` and \`[]engine.Glyph\`. For that ptrcast to be safe-by-spec, both element types must have guaranteed layout — i.e. \`extern struct\`.

The codegen \`@ptrCast\` lands in a sibling labelle-assembler PR.

## Why now

labelle-assembler#108 (sokol Phase 4 font + audio impl) flagged this as a real concern when reviewing its codegen interaction. Until #108, no concrete backend had implemented the font traits, so the slice-marshal codegen had never actually compiled. Now it has to.

## Test plan

- [x] \`zig build test\` — 373/373 passing.
- [x] No behavioural change — \`extern struct\` only restricts field ordering, doesn't change size or alignment for these all-primitive structs. Existing size-stability test in \`test/font_types_test.zig\` (locks Glyph=20, CodepointEntry=8, KernPair=12) continues to pass.

## References

- labelle-assembler#108 — sokol Phase 4 impl that flagged this
- labelle-engine#447 / #448 — Phase 4 tracking
- labelle-engine#529 — original font_types

### DIFF
--- a/src/font_types.zig
+++ b/src/font_types.zig
@@ -25,7 +25,15 @@ pub const FontId = struct {
 /// already incorporate the glyph's bearing — the renderer just adds
 /// them to the pen position. `advance` moves the pen for the next
 /// glyph. See `RFC-FONT-LOADER.md` §3.
-pub const Glyph = struct {
+///
+/// `extern struct` so the assembler-generated `FontBackendAdapter` can
+/// `@ptrCast` slices between `[]BackendGfx.Glyph` and `[]engine.Glyph`
+/// (and the equivalent sokol-side types) — the three repos define
+/// structurally-identical-but-nominally-distinct `Glyph` types and
+/// rely on a zero-cost reinterpret at the codegen marshal boundary.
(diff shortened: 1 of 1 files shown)


apotema:
@copilot review

Issue #536, opened 2026-05-13
Preview Phase 3: wire emitNodeEntered into the flow runtime


apotema:
Follow-up to #535. The `Preview.emitNodeEntered(flow_name, node_id)` primitive + `subscribe_flow` / `unsubscribe_flow` control-plane support landed in the PR for #535. The runtime call-site was **not** wired because the Flows runtime referenced in #535 ("merged sub-issues #43-#56") is not present in this repo on `main` as of that PR — there is no `Flow` type, no `flow.tick()`, no `.flow.zon` loader, and no per-node `enter()` dispatch to hook into.

## Scope

When the Flows runtime lands (separate umbrella), add a single call at the per-node dispatch site:

```zig
if (game.preview) |*p| p.emitNodeEntered(flow_name, node.id) catch {};
```

- `flow_name` is the `.flow.zon` file stem (no path, no extension).
- `node.id` is the stable u32 the flow_io parser assigns each node.
- The call is a no-op when `subscribed_flows` is empty (default), so the hot path stays cheap.

## Acceptance

- Integration test: load a flow, subscribe via `{"kind":"subscribe_flow","flow":"<stem>"}`, tick the runtime across a node boundary, assert one `node_entered` binary frame with the correct `(flow_name, node_id)`.
- No regression in #535's existing tests (emitter + opt-in subscription).

## References

- #535 — Phase 3 emitter (primitive shipped; this issue covers the runtime wire-up)
- labelle-gui#59 — PIE preview umbrella

apotema:
Misplaced — Flows compile to ordinary Zig scripts via labelle-gui's projector codegen (`labelle-gui/src/flows/projector.zig`, shipped via labelle-gui#50). There's no separate flow runtime in labelle-engine to hook into. The emitNodeEntered call needs to be inserted by the codegen at each node's emitted body, not at a runtime call site.

Refiled as the equivalent labelle-gui codegen issue. The engine primitive `Preview.emitNodeEntered` (shipped in #537) is ready to be called from the generated Zig.

Issue #518, opened 2026-05-12
Preview mode Phase 2: binary state telemetry frames


apotema:
Phase 2 of the PIE preview umbrella ([labelle-gui#59](https://github.com/labelle-toolkit/labelle-gui/issues/59)). Phase 1 (engine spike via #517) shipped hello/heartbeat/bye control plane. Phase 2 adds the **telemetry hot path**: binary frames that stream entity / component state from the running engine back to the editor so the editor can show what's happening at scale.

## Scope

Engine side only. Editor consumes via the existing TCP channel from #517.

### New frame types (length-prefixed binary)

| Kind | Direction | Payload |
|---|---|---|
| `entity_created` | engine → editor | entity_id (u64), prefab_name (optional, length-prefixed) |
| `entity_destroyed` | engine → editor | entity_id |
| `component_changed` | engine → editor | entity_id, component_name, raw bytes |
| `subscribe` | editor → engine | list of component names to stream |
| `unsubscribe` | editor → engine | list of component names |

### Frame envelope

```
[u8 magic] [u8 kind] [u32 length] [payload bytes...]
```

Mixed with the existing newline-delimited JSON control plane on the same TCP socket: a frame starting with magic byte `0x1B` (ESC) is binary; anything else is JSON. Keeps single-channel simple.

### Throttling

- Engine batches `component_changed` events per frame, sends one digest message at end-of-tick rather than one event per component touch.
- Editor subscribes to specific (component_name, entity_id) tuples to scope traffic.

## Out of scope

- Reverse channel from editor → engine for input forwarding (Phase 3+).
- Selection-sync messages (Phase 3 of #59).
- Pin-value telemetry for Flows visualization (Phase 4 of #59, separate work).
- Hot reload of files (separate Phase 2+ thread, not telemetry).

## Acceptance

- Add `tickTelemetry` to the existing `Preview` API in `labelle-engine/src/preview_mode.zig`.
- Three new frame emitters (`emitEntityCreated`, `emitEntityDestroyed`, `emitComponentChanged`) and a subscribe/unsubscribe state machine.
- Engine-side test that drives an ECS world through entity-lifecycle events with a connected `Preview` and asserts the emitted byte stream matches expectations.
- Manual smoke: connect a `nc -l` listener, run a game with `--preview-mode`, verify the binary frames appear interleaved with heartbeats.

## References

[...]

Issue #56, opened 2025-12-21
Expose Audio API for sound effects and music playback


apotema:
## Problem

There's no way to play audio (sound effects or music) from scripts in labelle-engine projects. The underlying raylib backend supports audio, but this functionality is not exposed through labelle-engine.

## Use Case

Games need audio for:
- Sound effects (e.g., button clicks, item pickups, character actions)
- Background music
- Ambient sounds

Example usage:
```zig
pub fn onCreate(entity: u64, game_ptr: *anyopaque) void {
    Audio.playSound("pickup.wav");
}

pub fn update(game: *Game, scene: *Scene, dt: f32) void {
    if (!Audio.isMusicPlaying()) {
        Audio.playMusic("background.ogg");
    }
}
```

## Proposed Solution

1. Add audio configuration to `project.labelle`:
```zon
.resources = .{
    .sounds = .{
        .{ .name = "pickup", .path = "resources/sounds/pickup.wav" },
    },
    .music = .{
        .{ .name = "background", .path = "resources/music/background.ogg" },
    },
},
```

2. Expose an Audio API through labelle-engine:
```zig
pub const Audio = struct {
    pub fn playSound(name: []const u8) void;
    pub fn playMusic(name: []const u8) void;
    pub fn stopMusic() void;
    pub fn setMusicVolume(volume: f32) void;
    pub fn setSoundVolume(volume: f32) void;
    pub fn isMusicPlaying() bool;
};
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #537, opened 2026-05-13
feat(preview): node_entered emitter (#535)

### LINKED ISSUE #535: Preview Phase 3: node_entered emitter from flow runtime
Phase 3 of the PIE preview umbrella ([labelle-gui#59](https://github.com/labelle-toolkit/labelle-gui/issues/59)). Engine-side primitive for **active-node highlight** — when the running game enters a flow node, the engine emits `node_entered(flow_id, node_id)` so the editor can pulse that node in the flow editor.

Sibling: #534 (watch_entity / selection sync), plus an editor-side consumer coordination issue in labelle-gui.

## Dependencies

Builds on the Flows infrastructure ([labelle-engine#42](https://github.com/labelle-toolkit/labelle-engine/issues/42) and the merged sub-issues #43–#56). The Flows runtime — wherever each node's `enter()` / `execute()` callback fires — is where this issue's emit hooks in.

## What this builds

Define a new binary telemetry frame `BinaryFrameKind.node_entered`:

```
[u8 magic=0x1B] [u8 kind=node_entered] [u32 length-LE] [payload]

payload:
    [u16 flow_name_len] [flow_name bytes (UTF-8)]
    [u32 node_id]
```

Why binary instead of JSON: node-enter events at runtime fire at potentially 60+ Hz across multiple flows. The Phase 2 telemetry plane already exists on the same socket for exactly this kind of high-frequency channel.

Engine emits `Preview.emitNodeEntered(flow_name, node_id)` from the flow-runtime's node-enter callback. Like the entity emitters, it's a no-op if `game.preview == null`, and gated by a new `subscribed_flows: std.StringHashMap(void)` (paralleling `subscribed_components`) so the editor can opt into per-flow firehose vs. blanket.

Control-plane subscribe frame:
```json
{"kind":"subscribe_flow","flow":"player_state_machine"}
{"kind":"unsubscribe_flow","flow":"player_state_machine"}
```

Empty `subscribed_flows` = no emit (default: editors that don't care pay zero cost). This is the inverse of `subscribed_components` (where empty = emit all) because flow nodes fire much more frequently than component changes — opt-in is safer.

## File layout

- `src/preview_mode.zig` — add `node_entered` to `BinaryFrameKind`; `emitNodeEntered` helper writing the binary frame; `subscribed_flows` map; `pollSubscription` recognizes `subscribe_flow` / `unsubscribe_flow` kinds.
- Flow runtime (wherever `node.enter()` is called from `flow.tick()` or equivalent) — call `if (game.preview) |*p| p.emitNodeEntered(flow_name, node.id) catch {};` at the right point in the dispatch. The `flow_name` here is the `.flow.zon` file stem; `node.id` is the stable u32 the flow_io parser already assigns.
- `test/preview_mode_test.zig` — mock-editor test: subscribe to flow `"test_flow"`, advance the runtime by one tick that crosses a node boundary, assert one `node_entered` binary frame with the correct flow name + node id.

## Acceptance

- `zig build test` green; new node_entered binary frame test passes.
- Subscribing to a non-existent flow name silently succeeds (editor may speculatively subscribe before the flow loads); no frames fire until the flow is loaded and that flow's node enters.
- Wire-format guard test: assert the exact byte layout of one `node_entered` frame matches the spec above.

## Out of scope

- Node-exit events (Phase 3+ if needed; the editor pulses a brief highlight on enter and doesn't need to track exit explicitly).
- Variable-watch within a node (deferred — labelle-gui#57 covers per-pin/edge value display, which is a separate consumer of `component_changed` once the flow runtime exposes intermediate values as components).
- Editor-side wiring (separate labelle-gui issue under the Phase 3 coordination).

## References

- labelle-gui#59 — PIE preview umbrella; Phase 3 description names this work as "see the flow running"
- labelle-engine#518 — Phase 2 binary frame design this extends
- labelle-engine#42 — Flows umbrella (runtime that this hooks into)

apotema:
## Summary

Phase 3 of the PIE preview umbrella (labelle-gui#59). Adds the engine-side primitive for the editor's "active flow node" highlight: a `node_entered(flow_name, node_id)` binary telemetry frame on the existing preview socket.

- New `BinaryFrameKind.node_entered = 4` (appended; discriminators 1-3 unchanged).
- New `subscribed_flows: StringHashMap(void)` with **opt-in** default (inverse of `subscribed_components`): empty set means no emit. Flow nodes fire at frame rate across many flows, so editors that don't care pay zero cost.
- `Preview.emitNodeEntered(flow_name, node_id)` writes `[u16 name_len LE][name][u32 node_id LE]` after the standard `[0x1B][kind][u32 len LE]` header. Early-out before allocation when not subscribed.
- `pollSubscription` now also decodes `{"kind":"subscribe_flow","flow":"..."}` / `{"kind":"unsubscribe_flow","flow":"..."}` control-plane frames. Caller-owned strings are duped into `subs_arena` to match the existing pattern.

## Flow-runtime hookup: deferred to #536

The hookup at the per-node dispatch site (`if (game.preview) |*p| p.emitNodeEntered(flow_name, node.id) catch {};`) is **not in this PR** because the Flows runtime referenced in #535's "Dependencies" section is not present on `main`: no `Flow` type, no `flow.tick()`, no `.flow.zon` loader, and no node-enter callback to hook into. The umbrella issue list (#43-#56) tracks runtime work that has not landed yet.

Once that infrastructure lands, the integration is a one-liner; #536 tracks it.

## Test plan

- [x] `zig build test` green locally (266 insertions, 9 deletions; 5 new tests).
- [x] Subscribed emit → asserts binary frame contents (kind, name_len, name bytes, node_id, tight payload sizing).
- [x] Unsubscribed default → no binary bytes precede a subsequent heartbeat (proves silence).
- [x] Subscribe + unsubscribe → emit after unsubscribe is silent (proves the membership flip works).
- [x] Hand-composed wire-format guard → exact byte-for-byte comparison of a `node_entered` frame against the spec in the issue.
- [x] Speculative subscribe to a not-yet-loaded flow name silently succeeds (matches acceptance criterion in #535).
- [x] Phase 1 / Phase 2 tests (hello, heartbeat, bye, entity_created, entity_destroyed, component_changed, malformed JSON) continue to pass.

## References

- Closes #535 (emitter + control-plane support; runtime hookup tracked in #536).
- Filed #536 for the runtime call-site once Flows lands.
- Extends #518 (Phase 2 binary frame design).
- labelle-gui#59 — PIE preview umbrella; Phase 3 names this work as "see the flow running".

### DIFF
--- a/src/preview_mode.zig
+++ b/src/preview_mode.zig
@@ -109,6 +109,7 @@ pub const BinaryFrameKind = enum(u8) {
     entity_created = 1,
     entity_destroyed = 2,
     component_changed = 3,
+    node_entered = 4,
     _,
 };
 
@@ -154,6 +155,13 @@ pub const Preview = struct {
     /// frames for. Keys allocated in `subs_arena` so the editor can
     /// churn the set without leaking into the per-frame `arena`.
     subscribed_components: std.StringHashMapUnmanaged(void) = .{},
+    /// Set of flow names (the `.flow.zon` file stem, no path / no
+    /// extension) the editor wants `node_entered` frames for. Inverse
+    /// default vs. `subscribed_components`: empty set means **no**
+    /// emit. Flow nodes fire much more frequently than component
+    /// changes (60+ Hz across many flows), so opt-in is the safer
+    /// default — editors that don't care pay zero cost.
+    subscribed_flows: std.StringHashMapUnmanaged(void) = .{},
     subs_arena: std.heap.ArenaAllocator,
     /// Inbound newline-framing buffer for `pollSubscription`. Owned
     /// by `inbox_alloc` (the parent allocator) so it can grow past the
@@ -186,6 +194,7 @@ pub const Preview = struct {
         // close here.
         self.stream.close();
         self.subscribed_components.deinit(self.subs_arena.allocator());
+        self.subscribed_flows.deinit(self.subs_arena.allocator());
         self.subs_arena.deinit();
         self.inbox.deinit(self.inbox_alloc);
         self.arena.deinit();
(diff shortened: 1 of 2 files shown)


apotema:
@copilot review

Issue #534, opened 2026-05-13
Preview Phase 3: watch_entity protocol + filtered component_changed emission


apotema:
Phase 3 of the PIE preview umbrella ([labelle-gui#59](https://github.com/labelle-toolkit/labelle-gui/issues/59)). Engine-side primitive for **selection sync** — the editor tells the engine which entity is selected, the engine streams component state changes for that entity (and only that entity) back.

Sibling work: a coordination issue in labelle-gui for the editor-side consumer (linked once filed), and a separate engine issue for the `node_entered` flow-runtime emitter.

## What's there today

- Phase 2 shipped `emit_component_changed(entity_id, component_name, raw_bytes)` (labelle-engine#518/#519) and wired it into `game.zig`'s ECS lifecycle (#520/#521).
- `Preview.pollSubscription` already reads `subscribe(components)` / `unsubscribe(components)` JSON frames and gates `component_changed` emission by component-name set.
- What's missing: **entity-id-based filtering** on top of component-name filtering. Today every entity's component-change flushes if the component is subscribed; Phase 3 wants only the *selected* entity's components.

## What this builds

Add two new control-plane JSON frames (editor → engine):

```json
{"kind":"watch_entity","id":42}
{"kind":"unwatch_entity","id":42}
```

Engine maintains a `watched_entities: std.AutoHashMap(u64, void)` on the `Preview` struct. `emitComponentChanged` checks BOTH:
1. The component name is in `subscribed_components` (existing Phase 2 check)
2. The entity id is in `watched_entities` (new Phase 3 check)

When `watched_entities` is empty, the emit fires for ALL entities (back-compat with Phase 2's "watch everything"). When non-empty, only watched entities fire. This lets the editor opt into entity-scoped telemetry without breaking the "show me everything" mode the Phase 2 binary tests rely on.

Plus a snapshot frame for the initial state after a watch:

```json
{"kind":"watch_entity","id":42}
// engine responds with a one-shot full snapshot of entity 42's subscribed components,
// then streams component_changed frames as they happen
```

The snapshot avoids the "stale UI until next mutation" problem when the editor watches a quiescent entity.

## File layout

- `src/preview_mode.zig` — add `watched_entities` field; `pollSubscription` extended to recognize `watch_entity` / `unwatch_entity` kinds; new `emitEntitySnapshot(entity_id, components)` helper.
[...]

Issue #520, opened 2026-05-12
Preview Phase 2 wiring: hook telemetry emitters into game.zig ECS lifecycle


apotema:
Phase 2 of the PIE preview umbrella ([labelle-gui#59](https://github.com/labelle-toolkit/labelle-gui/issues/59)). Engine #518 / PR #519 shipped the `emitEntityCreated` / `emitEntityDestroyed` / `emitComponentChanged` API on `Preview`, but **`game.zig` doesn't actually call them yet**. The spike deliberately deferred game-loop integration to a sibling sub-issue — this is that issue.

## What this builds

Wire the three Phase 2 emitters into `game.zig`'s ECS lifecycle paths:

- **Entity creation** — wherever `game.create_entity` / equivalent runs, emit `entity_created(id, prefab_name)` if `game.preview` is set.
- **Entity destruction** — same shape for `entity_destroyed(id)` on destroy paths (component-attached, manual, scene-unload).
- **Component changes** — emit `component_changed(id, comp_name, raw_bytes)` from the typed component setters / `addComponent` paths. Bounded by the `Preview.subscribed` filter set (engine already implements subscription state); skip the emit if the component name isn't subscribed.

Guard every call with `if (game.preview) |*p| { p.emit... catch {}; }` so production runs without preview pay nothing.

## Acceptance

- New end-to-end test in `test/preview_mode_test.zig` that:
  1. Stands up a loopback TCP listener (mock editor).
  2. Spawns a game with `--preview-mode 127.0.0.1:<port>`.
  3. Walks through the editor protocol: receive `hello` + heartbeat.
  4. Sends `subscribe([Position, Sprite])` to the engine.
  5. Game creates an entity with Position + Sprite components.
  6. Listener receives `entity_created` + two `component_changed` frames.
  7. Game destroys the entity → listener receives `entity_destroyed`.
  8. Game changes Position without subscribing first → no frame emitted.
- `zig build` clean across all backends.
- `zig build test` green.

## Out of scope

- Editor-side consumer (an "Entities" panel in labelle-gui showing live state). That's a separate `labelle-gui` Phase 2 issue under #59.
- Per-tick batching (the engine module has a `TODO` for it — not adding here).
- Selection sync messages (Phase 3 of #59).
- Pin-value telemetry for Flows visualization (Phase 4 of #59).

## References

[...]

PR #538, opened 2026-05-13
feat(preview): watch_entity protocol + filtered component_changed (#534)

apotema:
## Summary

Phase 3 of the PIE preview umbrella (labelle-gui#59). Adds entity-id scoped filtering on top of the Phase 2 component-name filter so the editor can opt into selection-sync ("stream just the selected entity") without breaking the Phase 2 "watch everything" contract.

- New `Preview.watched_entities: AutoHashMapUnmanaged(u64, void)` plus `isEntityWatched` policy chokepoint.
- `pollSubscription` learns two new JSON kinds: `{"kind":"watch_entity","id":42}` and `{"kind":"unwatch_entity","id":42}`. Existing `subscribe` / `unsubscribe` shapes are untouched.
- `emitComponentChanged` now applies the entity filter alongside the existing component-name filter. The Phase 2 binary frame layout is unchanged — the new behaviour is purely an additional gate on emission.
- New `emitEntitySnapshot(id, components)` helper writes N back-to-back `component_changed` frames so the editor doesn't sit on a stale view until the next mutation. Bypasses the entity filter (caller has already decided the entity is interesting) but honours the component-name subscription set.

## Design decision: empty-set semantics

When `watched_entities.count() == 0` the engine falls back to Phase 2 "watch everything" — every entity whose component is subscribed emits. The alternative ("empty set = strict include list, nothing flows") was rejected because:

1. It would break the Phase 2 binary-frame tests and the implicit contract that `subscribed_components` alone is enough to start receiving telemetry.
2. Today's editor watches one entity at a time; an empty set after unwatch will most commonly mean "user deselected", and going silent would surprise callers that rely on the global stream.
3. Multi-entity tracking is Phase 4+; the empty-set rule lives in a single chokepoint (`isEntityWatched`) so flipping the semantics later is a one-line change.

The decision is exercised by the `unwatch_entity` test which asserts the post-unwatch stream resumes for unrelated entities.

## Scope split

The `game.zig` side wiring — translating an inbound `watch_entity` into a walk of the ECS that feeds `emitEntitySnapshot` — is deliberately left to a follow-up issue. The helper's docstring documents the gap. This PR ships the wire-format primitive with full test coverage.

## Test plan

- [x] `zig build test` green.
- [x] Test count delta: **+5** (373 -> 378).
- [x] New tests in `test/preview_mode_test.zig`:
  - `watch_entity: adds id to watched_entities and filters subsequent emits`
  - `unwatch_entity: empty set restores Phase 2 'watch everything' behaviour`
  - `emitEntitySnapshot: writes one component_changed frame per (entity, component)`
  - `emitEntitySnapshot: skips components the editor did not subscribe to`
  - `emitEntitySnapshot: bypasses watched_entities filter for unwatched ids`
- [x] Phase 2 lifecycle test (`#520` wiring) still passes — back-compat verified.

Refs #534, #84.

### DIFF
--- a/src/preview_mode.zig
+++ b/src/preview_mode.zig
@@ -31,6 +31,8 @@
 ///        {"kind":"bye","reason":"normal"}\n
 ///        {"kind":"subscribe","components":["Position","Velocity"]}\n   (editor → engine)
 ///        {"kind":"unsubscribe","components":["Velocity"]}\n             (editor → engine)
+///        {"kind":"watch_entity","id":42}\n                               (editor → engine, Phase 3)
+///        {"kind":"unwatch_entity","id":42}\n                             (editor → engine, Phase 3)
 ///
 /// 2. **Binary telemetry plane** (Phase 2) — length-prefixed records
 ///    led by an `ESC` (0x1B) magic byte:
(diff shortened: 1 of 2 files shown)


apotema:
@copilot review

PR #539, opened 2026-05-14
feat: migrate to Zig 0.16

apotema:
## Summary
Partial 0.16 migration — library compiles, tests need follow-up.

- Bumps `minimum_zig_version` to 0.16.0
- `@Type(.{ .@"enum" })` → `@Enum(...)` (animation_def.zig)
- `@Type(.{ .@"fn" })` → `@Fn(...)` (query.zig — `separateComponents` marked `inline`)
- `@Type(.{ .@"struct" })` → `@Struct(...)` (script_runner.zig)

## Status
- `zig build` (lib): ✅ PASS
- `zig build test`: ❌ FAIL (~80 errors from 0.16 `Io`-everywhere stdlib redesign)

## Known follow-ups
Test failures are from Zig 0.16's stdlib reshuffle of net/thread/timer/dir to require `Io` threading:
- `std.net` → `std.Io.net`: `src/preview_mode.zig` TCP module (~300 lines) needs Io-Stream/Server rewrite
- `std.Thread.{sleep,Mutex}` + `std.time.Timer` moved into `Io`: assets/{catalog,worker}.zig, jsonc/deserializer.zig, script_runner.zig
- `std.Io.Dir` redesign: ~6 test files use removed `makeDir`/`writeFile` shapes

Recommend splitting these into focused follow-up PRs per subsystem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install Zig
         uses: mlugg/setup-zig@v2
         with:
-          version: 0.15.2
+          version: 0.16.0
 
       - name: Cache Zig dependencies
         uses: actions/cache@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
       - name: Checkout sibling dependencies
         run: |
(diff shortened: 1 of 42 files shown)


apotema:
@copilot review

PR #540, opened 2026-05-14
feat(preview): pin_value binary frame + subscribe_pin_values control kind

apotema:
## Summary

Engine-side primitive for the Flows live pin/edge value preview tracked
in `labelle-toolkit/labelle-gui#57`. Adds the Phase 3 binary telemetry
channel for per-pin runtime values, structured as a literal mirror of
the `node_entered` emitter shipped in #537.

- New `BinaryFrameKind.pin_value = 5` (appended; existing 1-4 wire-stable).
- New `Preview.emitPinValue(flow_name, node_id, pin_name, value: f64)`
  with an opt-in `subscribed_pin_flows` set — empty set = no emit.
- `pollSubscription` decodes `subscribe_pin_values` /
  `unsubscribe_pin_values` JSON control frames.
- Scope is **just the engine primitive**. Editor consumer and codegen
  instrumentation are deferred follow-ups (see below).

## Wire format

Standard binary header followed by the new payload:

```
[u8 magic=0x1B] [u8 kind=5] [u32 length LE]    // header (6 bytes)
[u16 flow_name_len LE] [flow_name bytes]       // payload start
[u32 node_id LE]
[u16 pin_name_len LE]   [pin_name bytes]
[f64 value LE]                                  // bits cast u64 → LE
```

Example: `emitPinValue(\"hud\", 0x01020304, \"x\", 1.0)` produces
20 payload bytes:

```
1B 05  14 00 00 00            // magic, kind=pin_value, len=20 LE
03 00  68 75 64                // flow_name_len=3, \"hud\"
04 03 02 01                    // node_id LE
01 00  78                      // pin_name_len=1, \"x\"
00 00 00 00 00 00 F0 3F        // f64 1.0 LE
```

The `emitPinValue: exact wire-format guard against drift` test
hand-composes this exact byte sequence and diffs the emitter's
output against it — any future drift will fail loud.

**v1 value is `f64`-only.** Covers BinOp results, Literal numerics,
and `dt`. Strings / bools / structs are deferred (see below).

## Subscription protocol

Mirrors `subscribe_flow` / `unsubscribe_flow` but tracks pin-value
emission independently:

```json
{\"kind\":\"subscribe_pin_values\",\"flow\":\"player_state_machine\"}
{\"kind\":\"unsubscribe_pin_values\",\"flow\":\"player_state_machine\"}
```

Why a separate kind (rather than folding pin-values into
`subscribe_flow`): node pulses and pin payloads are very different
volume profiles. A future minimap-style editor may want node pulses
without paying for the high-frequency pin stream; a future debugger
may want pins without the node-pulse channel. Keeping the sets
independent preserves symmetry with `node_entered` and avoids a
flag-bit hack inside one control message. The editor will simply
subscribe to both when opening a flow tab.

Caller-owned strings are duped into the existing `subs_arena` — same
pattern `subscribe_flow` uses, so the editor can churn the set
without leaking into the per-frame arena.

## Tests

8 new tests in `test/preview_mode_test.zig`:

1. `emitPinValue: no-op when flow is not pin-subscribed` — opt-in default.
2. `emitPinValue: subscribed flow emits a frame with the expected layout` —
   decode every field, assert tight payload size.
3. `emitPinValue: unsubscribe_pin_values stops emission`.
4. `emitPinValue: subscribing to flow A doesn't enable emission for flow B`.
5. `emitPinValue: pin_name with non-ASCII and quotes round-trips through
   the wire` — quotes, ümlauts, CJK, tabs/newlines.
6. `emitPinValue: f64 value preserves precision through encode/decode` —
   pi, e, ±floatMax, floatMin, floatEps, 1/3, ±0.0; bit-exact compare.
7. `emitPinValue: exact wire-format guard against drift` — hand-composed
   bytes diffed byte-for-byte.
8. `pollSubscription: subscribe_pin_values / unsubscribe_pin_values
   update subscribed_pin_flows` — and asserts independence from
   `subscribed_flows`.

`zig build` and `zig build test` are both green.

## What's deferred

This PR ships ONE piece of `labelle-gui#57`. Three follow-ups need to
land before the umbrella is done:

- **Non-`f64` value types.** v1 is `f64`-only (BinOp results, Literal
  numerics, `dt`). String / bool / struct payloads need a value-type
  tag in the payload (e.g. `[u8 value_kind] [variable-shape value]`).
  Should be filed as a follow-up issue on this repo before merge —
  the wire format will need a backwards-compat strategy (either a
  new `BinaryFrameKind.pin_value_typed = 6` or a v2 of this kind).
- **Editor consumer.** `labelle-gui` needs to subscribe on flow-tab
  open, decode `pin_value` frames, and render values on edges. Lives
  in `labelle-gui#57` proper.
- **Codegen instrumentation.** `flow_codegen` needs to emit
  `if (game.preview) |*p| p.emitPinValue(...)` calls at edge
  evaluation sites. Likely a new issue on this repo (parallel to the
  `#536` follow-up that wires `emitNodeEntered` into the per-node
  dispatch site).

Part of `labelle-toolkit/labelle-gui#57`.

### DIFF
--- a/src/preview_mode.zig
+++ b/src/preview_mode.zig
@@ -112,6 +112,7 @@ pub const BinaryFrameKind = enum(u8) {
     entity_destroyed = 2,
     component_changed = 3,
     node_entered = 4,
+    pin_value = 5,
     _,
 };
 
@@ -177,6 +178,13 @@ pub const Preview = struct {
     /// changes (60+ Hz across many flows), so opt-in is the safer
     /// default — editors that don't care pay zero cost.
     subscribed_flows: std.StringHashMapUnmanaged(void) = .{},
+    /// Set of flow names the editor wants `pin_value` frames for.
+    /// Mirror of `subscribed_flows` but tracks the (potentially much
+    /// higher-volume) live pin/edge value channel separately so the
+    /// editor can opt into node pulses without paying for per-pin
+    /// payloads (or vice versa). Empty set means **no** emit — same
+    /// opt-in semantics as `subscribed_flows`.
+    subscribed_pin_flows: std.StringHashMapUnmanaged(void) = .{},
     subs_arena: std.heap.ArenaAllocator,
     /// Set of entity IDs the editor wants `component_changed` frames
     /// scoped to. See the field-level doc on `Preview` for the
@@ -214,6 +222,7 @@ pub const Preview = struct {
         self.stream.close();
         self.subscribed_components.deinit(self.subs_arena.allocator());
         self.subscribed_flows.deinit(self.subs_arena.allocator());
+        self.subscribed_pin_flows.deinit(self.subs_arena.allocator());
         self.watched_entities.deinit(self.subs_arena.allocator());
         self.subs_arena.deinit();
         self.inbox.deinit(self.inbox_alloc);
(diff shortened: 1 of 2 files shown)


apotema:
@copilot review

PR #541, opened 2026-05-14
feat(game): getComponent + setField + preview accessor for flow codegen

apotema:
## Summary

Lands the engine slice of labelle-toolkit/labelle-gui#96 — the API surface the flow-codegen in `labelle-gui/flow-codegen/src/codegen.zig` assumes when it emits node bodies. Two methods (one re-ordered, one new) on `Game` plus a regression test pinning down the existing `preview` field shape.

- `Game.getComponent` now takes `(comptime T, entity)` so `game.getComponent(Position, entity)` compiles unchanged.
- `Game.setField` is new: writes a single comptime-selected field of component `T` on an entity, no-ops silently when the entity lacks the component, and routes through `notifyComponentChanged` so preview subscribers see the mutation.
- `game.preview` was already an optional-by-value `Preview` field (issue #537); the codegen's `if (game.preview) |*_p|` capture works as-is, and a new test pins it down.

## API

```zig
pub fn getComponent(self: *Self, comptime T: type, entity: Entity) ?*T

pub fn setField(
    self: *Self,
    comptime T: type,
    comptime field: std.meta.FieldEnum(T),
    entity: Entity,
    value: @FieldType(T, @tagName(field)),
) void

// already on Game (#537):
preview: ?preview_mode_mod.Preview = null
```

Call sites the codegen emits — verified by inspection against `labelle-gui/flow-codegen/src/codegen.zig` and by tests in `test/flows_game_api_test.zig`:

```zig
const n0_value = game.getComponent(Position, entity) orelse return;
game.setField(Position, .x, entity, n4_value);
if (game.preview) |*_p| {
    _p.emitNodeEntered("flow_name", 1) catch {};
}
```

## Decisions

- **Parameter order on `getComponent`** — the old signature was `(entity, comptime T)`. Swapped to `(comptime T, entity)` to match the codegen instead of forking the codegen template; in-tree only had four test files calling it and they were updated in this PR. The backend primitive `ecs_backend.getComponent(entity, T)` is unchanged.
- **`?*T` over `?T`** — kept the pointer shape because the existing ECS-backend primitive (MockEcsBackend, zig_ecs, zflecs all matching) returns `?*T`, and codegen's `orelse return` works either way. Pointer aliases storage, which is the natural ECS contract.
- **No-op vs assert on missing component in `setField`** — silent no-op. Justified by the codegen pattern: a generated `OnUpdate` flow that touches a component the entity dropped this frame should not crash the game loop. Documented on the public function.
- **`setField` fires `notifyComponentChanged`** — matches `setComponent`'s contract so subscribed editors still see field-level mutations as `component_changed` frames.
- **`@FieldType` builtin (not `std.meta.FieldType`)** — Zig 0.15.2 removed `std.meta.FieldType`; the builtin is the supported path.

## Test plan

New file `test/flows_game_api_test.zig` with 7 tests, all green under `zig build test` (390/390 total):

- [x] `getComponent` returns null for missing component
- [x] `getComponent` returns a pointer aliasing storage
- [x] `setField` updates the named field in place; sibling field untouched
- [x] `setField` on entity without component is a silent no-op
- [x] `setField` accepts the exact codegen-emitted call shape `game.setField(T, .field, entity, value)`
- [x] `game.preview` is null on a default-constructed Game
- [x] `if (game.preview) |*_p| _p.emitNodeEntered(...)` resolves end-to-end against a loopback `Preview.connect`

Plus four in-tree test files updated to the new `getComponent(T, entity)` param order:

- `test/root_test.zig`
- `test/set_sprite_flip_test.zig`
- `test/jsonc_bridge_gizmo_visibility_test.zig` (sed-style bulk swap, all sites)
- (and `build.zig` lists the new test file)

Part of labelle-toolkit/labelle-gui#96

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -81,6 +81,7 @@ pub fn build(b: *std.Build) void {
         "test/collect_entities_test.zig",
         "test/set_sprite_flip_test.zig",
         "test/preview_mode_test.zig",
+        "test/flows_game_api_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/src/game.zig
+++ b/src/game.zig
@@ -1055,6 +1055,34 @@ pub fn GameConfig(
             return self.ecs_backend.getComponent(entity, T);
         }
 
+        /// Writes a single field of component `T` on `entity` in place.
+        ///
+        /// Silently no-ops when the entity does not have a component
+        /// of type `T` — same semantics as `getComponent` returning
+        /// `null`. This matches the flow-codegen runtime contract:
+        /// generated `OnUpdate` flows that touch a missing component
+        /// should not crash the game loop.
+        ///
+        /// The `comptime field: std.meta.FieldEnum(T)` selector keeps
+        /// the call site type-checked end-to-end. The flow-codegen
+        /// emits `game.setField(Position, .x, entity, value)`.
+        ///
+        /// Preview telemetry is fired after the mutation through the
+        /// same `notifyComponentChanged` path used by `setComponent`,
+        /// so the editor sees an updated component frame when it has
+        /// subscribed to `T`.
+        pub fn setField(
+            self: *Self,
+            comptime T: type,
+            comptime field: std.meta.FieldEnum(T),
+            entity: Entity,
+            value: @FieldType(T, @tagName(field)),
+        ) void {
+            const comp_ptr = self.ecs_backend.getComponent(entity, T) orelse return;
+            @field(comp_ptr.*, @tagName(field)) = value;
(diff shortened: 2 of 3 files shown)


apotema:
@copilot review

apotema:
Reworked: removed the `getComponent` parameter reorder after counting **48 external call sites** of the existing `(entity, T)` order across labelle-cli, labelle-gfx, bakery-game, and downstream consumers. Engine API is now unchanged on that surface — only addition is `setField`. The codegen catches up to the engine via labelle-gui#97 instead.

PR #558, opened 2026-05-20
feat(android): runtime immersive mode (hide system bars via JNI)

apotema:


apotema:
Companion assembler PR (emits the `engine.android.enableImmersiveMode()` call into the generated Android `main.zig`): labelle-toolkit/labelle-assembler#153

Issue #570, opened 2026-05-20
Audit scenes//prefabs/ name collisions before unifying the registry scan


apotema:
Part of #560 (RFC: unify scenes and prefabs).

## Problem

#561's loader change adds `scenes/` to the registry scan and merges both directories into one flat name-keyed registry. The moment that lands, any `scenes/foo.jsonc` + `prefabs/foo.jsonc` pair — plausible: a scene named after its primary prefab — becomes an **instant load-time collision**, independent of the file-format migration. This can break in-tree projects (flying-platform-labelle) before a single file is migrated.

## Scope

- Audit every in-tree project (`flying-platform-labelle`, examples, test fixtures) for basename collisions across `scenes/` and `prefabs/`, including nested folders.
- For each collision: rename, add an explicit `"name"`, or apply the path-qualified-reference form (if that resolution strategy lands).
- Land the resolutions **before or with** #561 so the merged scan never lands red.

## Acceptance

- Zero basename collisions across `scenes/` + `prefabs/` in every in-tree project.
- #561 can be merged without breaking any project's load.

Refs: #560, #561.

Issue #562, opened 2026-05-20
RFC §unification: spec overrides merge rules


apotema:
Part of #560.

Specify the exact merge semantics for `"overrides"` on reference entries. The unification doesn't change today's runtime behavior; this ticket is to write the rules down so the loader and authoring tools (humans + LLMs) can rely on them.

Decisions to lock in:

- **Deep vs shallow merge** for components mentioned in overrides.
- **List semantics** inside a component value — replace, append, per-field-typed?
- **Component present in referenced prefab, omitted in overrides** — kept (today's behavior).
- **Component present in overrides, not in referenced prefab** — added.
- **Removing a component** the referenced prefab has — proposed syntax: `"Position": null`. Decide.

Document the rules in the RFC or a follow-up doc once decided.

Issue #560, opened 2026-05-20
RFC: unify scenes and prefabs


apotema:
## Summary

Unify the scene and prefab file formats into one. Every `.jsonc` under `scenes/` or `prefabs/` follows the same shape; the engine builds one flat name-keyed registry across both directories. Folders become organizational, not semantic.

## Motivation

Scenes and prefabs share ~90% of their grammar but live in two code paths with two sets of assumptions. The asymmetry has surfaced as a recurring shape of bug:

- #467 — nested scene entities didn't fire `postLoad`
- #470 — `Parent` wasn't persisted for scene-rooted entities
- `flying-platform-labelle#286` — decor children disappeared after F9

Each was a different symptom of "the engine treated scene entities differently from prefab entities, by accident." The structural fix is removing the asymmetry.

## Proposal at a glance

- Explicit `"root"` wrapper for entity content; file metadata (`name`, future `version`, etc.) sits beside it at the file top.
- Child entries split into two disjoint modes: inline (`components` + `children`) vs reference (`prefab` + `overrides`).
- Optional `"name"` field; defaults to filename basename; collisions on the effective name = load-time error.
- Implicit/explicit root entity for every prefab (scenes today have none — they gain one). Unifies lifecycle (`destroy root → Parent cascade`), save/load, and composability.
- Scene-only behaviors find new homes: assets inferred from sprite refs (with `AssetManifest` component as escape hatch), default camera inserted by the engine at root instantiation only, state binding stays in `project.labelle`.
- Embedded entity defs inside component data (`Room.movement_nodes` pattern) preserved as-is; the `components` → `overrides` rename applies wherever a prefab reference appears.

## Open questions

- `.initial_scene` rename (lean: `.initial_prefab`).
- Warn-vs-silent when `"name"` diverges from basename.
- Exact `overrides` merge rules (deep vs shallow, list semantics, component removal syntax).
- Asset inference completeness audit (scripts that reference atlases never seen by inference).
- `AssetManifest` lifetime / refcount semantics on prefab destruction.

## Document

Full text on the branch attached to this issue: `RFC-UNIFY-SCENES-AND-PREFABS.md` at the repo root.

## Sub-tickets

- [ ] #561 Loader: accept unified prefab/scene format (with legacy synonyms)
- [ ] #562 Spec `overrides` merge rules
- [ ] #563 Sprite-based asset inference + `AssetManifest` component
- [ ] #564 Default `Camera` at root-only instantiation
[...]

PR #573, opened 2026-05-21
Loader accepts the unified prefab/scene format (#561)

### LINKED ISSUE #561: RFC §unification: loader accepts unified prefab/scene format
Part of #560.

Implement loader-side support for the unified prefab/scene format:

- Scan both `prefabs/` and `scenes/` into one flat name-keyed registry.
- Effective name = file's `"name"` field if present, else filename basename.
- Collision on effective name = load-time error.
- Accept the explicit `"root"` wrapper at file top.
- Accept `"overrides"` keyword on reference entries (split from `"components"`).

Legacy synonyms during the migration window — accept with a deprecation warning:

- `"entities"` → `"children"` at file root.
- `"assets"` → drop (warn).
- `"components"` on a reference entry (one with `"prefab"`) → `"overrides"`.

Plan to remove legacy synonyms after all in-tree files migrated.

Blocks on the `overrides` merge rules ticket — pickup once those are written down.

apotema:
Implements the loader side of the unified prefab/scene format — ticket #561, part of the unify-scenes-and-prefabs RFC (#560). Targets the umbrella branch `feat/unify-scenes-prefabs`.

## What's in

**Format acceptance + legacy synonyms** (`f2eba54`)
- New `src/jsonc/unified_format.zig` — an accessor layer that bridges the legacy and unified file shapes so the loader walks both down one code path.
- `rootObject` unwraps the explicit `"root"` wrapper; `fileChildren` reads `root.children` or legacy top-level `"entities"`; `entityPatch` reads `"overrides"` on a prefab reference or `"components"` on an inline entity.
- Legacy spellings still load with a one-time deprecation warning: `"entities"`, `"components"` on a reference, and a dropped `"assets"` field.
- Wired into `loadSceneFile`, `loadSceneSource`, `loadEntityInternal`, `spawnPrefabImpl`, `spawnAndLinkNestedEntities`.

**Effective-name resolution + collision detection** (`27d569c`)
- `unified_format.effectiveName` — a file's registry key is its `"name"` field, else the basename.
- `addEmbeddedPrefab` keys the cache by effective name; a duplicate is a load-time error (`error.DuplicatePrefabName`). Covers the production/mobile/WASM path with no assembler changes.

## Tests

12 new tests in `test/jsonc/unified_format_test.zig` — `root` wrapper, nested children, legacy `entities`/`assets`, `overrides` vs legacy `components`, cross-compatibility both directions (unified scene ↔ legacy prefab), effective-name resolution, and collision errors. **449/449 engine tests pass.**

## Scoped out (follow-up)

- **Eager filesystem registry scan** of `scenes/`+`prefabs/` — gated on the #570 collision audit and on the assembler emitting scene sources into the unified registry. Tracked separately.
- **`overrides` merge semantics** — #562's spec work. This PR keeps today's per-component override behavior unchanged; the keyword rename does not change behavior, per the RFC.

Closes #561 partially — the eager scan bullet stays open via the follow-up.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -98,6 +98,7 @@ pub fn build(b: *std.Build) void {
         "test/example_prefab_animation_walkthrough_test.zig",
         "test/jsonc/bridge_deserialize_test.zig",
         "test/jsonc/deserializer_test.zig",
+        "test/jsonc/unified_format_test.zig",
         "test/jsonc_bridge_gizmo_visibility_test.zig",
         "test/collect_entities_test.zig",
         "test/set_sprite_flip_test.zig",
--- a/build.zig
+++ b/build.zig
@@ -224,24 +224,31 @@ pub fn build(b: *std.Build) void {
     // `describe`-style nested structs rather than flat `test` blocks.
     // Wired into `test_step` as well so `zig build test` (what CI
     // runs) stays the single command that exercises everything.
-    const zspec_dep = b.dependency("zspec", .{ .target = target, .optimize = optimize });
-    const spec_module = b.createModule(.{
-        .root_source_file = b.path("spec/spec_tests.zig"),
-        .target = target,
-        .optimize = optimize,
-        .imports = &.{
-            .{ .name = "zspec", .module = zspec_dep.module("zspec") },
-            .{ .name = "labelle-core", .module = core_module },
-            .{ .name = "engine", .module = engine_module },
-            .{ .name = "scene", .module = scene_module },
-        },
-    });
-    const spec_tests = b.addTest(.{
-        .root_module = spec_module,
-        .test_runner = .{ .path = zspec_dep.path("src/runner.zig"), .mode = .simple },
-    });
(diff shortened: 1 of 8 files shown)


apotema:

BLOCKER
- None.

MAJOR
- `src/jsonc/unified_format.zig:30` uses `std.heap.page_allocator` for warning state, reintroducing the emscripten memory-growth hazard this codebase avoids with `c_allocator`.

MINOR
- `build.zig:227` resolves remote `zspec` unconditionally, so non-test builds/downstream consumers now require a test-only dependency to be available.
- `src/jsonc/scene_loader.zig:152` applies effective-name registration only to embedded prefabs; filesystem fallback still cannot resolve prefabs by their declared `"name"`.

NIT
- None.

Overall verdict: The unified-format loader path is well covered and `zig build` / `zig build test` pass locally, but I would address the allocator issue before merge because it conflicts with existing WASM/emscripten safety conventions. The other two items are scope/build-hygiene risks rather than immediate correctness blockers.

PR #574, opened 2026-05-21
Deep-merge semantics for prefab overrides (#562)

### LINKED ISSUE #562: RFC §unification: spec overrides merge rules

apotema:
Specifies and implements the merge rules for `overrides` on a prefab reference — resolves open question 3 of the unify-scenes-and-prefabs RFC (#560), ticket #562.

**Stacked on #573** (base `feat/561-unified-loader`) — #562 builds on #561's `overrides` keyword. Review/merge #573 first; this PR's base retargets to the umbrella branch once #573 lands.

## Decisions (confirmed with the project owner)

- **Deep field-level merge** — a field the override omits keeps the prefab's value, recursing into nested struct fields. This changes the pre-#562 behavior, where a partial override silently reset unmentioned fields to their struct defaults (the footgun the RFC's open question called out).
- **Arrays and scalars replace** outright — no element-wise list merge.
- **`"Name": null` removes a component** from the instance; siblings untouched; the removed component fires no `onReady`/`postLoad`.

## What's in

- `RFC-OVERRIDES-MERGE-RULES.md` — the spec doc, per-case with examples.
- `unified_format.mergeValues` / `mergedOverride` — recursive merge over JSONC `Value` trees. Merged trees are arena-scoped with leaf strings shared from the inputs, so deferred `@ref` names collected from a merged component stay valid into pass-2 resolution.
- `scene_loader.zig` — `loadEntityInternal` and `spawnAndLinkNestedEntities` merge per component and skip a `null` override (recorded in `applied` so the prefab fallback skips it too). A deep-merged component keeps its prefab's entity-bearing fields, so the nested-entity spawn pass uses the merged value.
- `on_ready.fireOnReadyAll` — skips a `null` override.

## Tests

6 zspec specs in `spec/override_merge_spec.zig` — field inheritance, nested-struct recursion, array replacement, component removal, sibling survival, and nested-entity survival through a merge.

**449 std.testing + 12 zspec tests pass.** The deep-merge behavior change regressed nothing — existing content uses full-component overrides (or `Position` with both fields).

Resolves #562.

### DIFF
--- a/RFC-OVERRIDES-MERGE-RULES.md
+++ b/RFC-OVERRIDES-MERGE-RULES.md
@@ -1,0 +1,127 @@
+# RFC Addendum: `overrides` Merge Rules
+
+**Status:** Accepted
+**Resolves:** open question 3 of `RFC-UNIFY-SCENES-AND-PREFABS.md` (#560)
+**Ticket:** #562
+
+## Scope
+
+In the unified prefab/scene format, a **reference entry** instantiates an
+existing prefab and may patch it:
+
+```jsonc
+{ "prefab": "hydroponics", "overrides": { "Health": { "current": 30 } } }
+```
+
+`overrides` is a map of component-name → patch. This document specifies
+exactly how that patch combines with the referenced prefab's components.
+
+The rules apply **everywhere a reference entry can appear** — in a
+`children` array, and inside an entity-bearing component field (the
+`Room.movement_nodes` pattern). They do **not** apply to the runtime
+`game.spawnFromPrefab(name, pos)` path, which takes no overrides.
+
+## The rule, in one line
+
+`overrides` is a **deep merge** onto the referenced prefab's component
+set: objects merge recursively, arrays and scalars replace, and a
+component whose patch value is `null` is removed.
+
+## Per-case specification
+
+Let *prefab* be the referenced prefab's components and *ov* be the
+`overrides` map.
+
+### 1. Component in *prefab* only
+
+Kept verbatim.
+
+### 2. Component in *ov* only
+
+Added to the instance, exactly as written.
+
+### 3. Component in both — deep merge
+
+The component value is merged field by field:
+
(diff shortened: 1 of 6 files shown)


apotema:

BLOCKER

- None.

MAJOR

- `src/jsonc/unified_format.zig:122` — On allocation failure, `mergeValues` silently falls back to whole-component replacement, violating the accepted deep-merge semantics and potentially dropping inherited fields under memory pressure.
- `src/jsonc/unified_format.zig:136` — A later append failure also returns the raw patch, so partially mergeable overrides can nondeterministically lose all prefab fields instead of surfacing `error.OutOfMemory`.

MINOR

- `src/jsonc/scene_loader.zig:389` — `null` removal semantics apply to all `entityPatch` results, including inline `components`, even though the RFC scopes component removal to `overrides`.
- `src/jsonc/on_ready.zig:48` — Same scope leak: inline `components: { "Name": null }` now suppresses lifecycle hooks as if it were an override removal.

NIT

- None.

Overall verdict: The core behavior is well-covered and the happy-path implementation matches the RFC, including nested entity-bearing fields and lifecycle suppression for removed components. I would not merge until the allocator-failure path is changed to propagate `OutOfMemory` instead of silently changing merge semantics; the inline-`components` null handling is less urgent but should be tightened or explicitly documented.

Issue #563, opened 2026-05-20
RFC §unification: sprite-based asset inference + AssetManifest component


apotema:
Part of #560.

Replace the scene-level `"assets"` declaration with engine-driven asset loading:

- **Inference.** Walk the instantiated prefab tree at load, collect `Sprite.sprite_name` references — including refs inside entity-bearing component fields like `Room.movement_nodes` / `Room.workstations`. Map to declaring resource bundles via `project.labelle`'s `.resources` and load before instantiation.
- **Escape hatch.** Introduce an `AssetManifest` component. Any prefab can carry `{ "AssetManifest": { "load": [...] } }` on its root (or any entity) to explicitly pre-load atlases / audio banks that inference can't see (script-loaded, runtime overlays, etc.).

Lifetime / refcount semantics deferred to the audit ticket (see #560 checklist).

apotema:
RFC updated with the pinned asset-inference design — see the new "Assets — inference + lazy fallback over \`AssetCatalog\`" section in \`RFC-UNIFY-SCENES-AND-PREFABS.md\`. Key resolved decisions for this ticket:

- **Foundation.** Built on the existing \`AssetCatalog\` (\`labelle-engine/src/assets/\`). Refcount, async decode, frame-budget upload, worker thread infrastructure all already exist. This ticket's scope is adding the **reverse index** and the **walker** on top.
- **Walker scope.** Walk every string in component data (including embedded refs inside component fields like \`Room.movement_nodes\`) against the reverse index. False positives are cheap (load an unneeded asset); the alternative (per-component-field declarations) puts the silent-missing-asset bug on the wrong side.
- **Two phases.** Phase A at engine startup builds the reverse index + per-prefab \`required_resources[]\`. Phase B at top-level spawn does \`acquire → wait → spawn → remember\`.
- **Reverse index shape.** Unified across atlas-sprites and standalone images:
  \`\`\`zig
  const ResourceRef = union(enum) {
      atlas: []const u8,
      image: []const u8,
  };
  \`\`\`
  So this ticket coordinates with #568 (Image component) — both consume the same index.
- **Collisions.** Load-time error at engine startup. Exact-match, case-sensitive. (Resolves an earlier open question.)
- **Lazy on-miss.** \`findSprite\` / \`findImage\` miss → \`assets.acquire(bundle)\` async; renderer skips the entity until \`isReady\`. Pop-in once loaded.
[...]

PR #576, opened 2026-05-21
Single shared entity-tree walker with cycle detection (#569)

### LINKED ISSUE #569: Single shared tree-walker for all entity-tree consumers
Part of #560 (RFC: unify scenes and prefabs).

## Problem

The RFC removes the scene/prefab asymmetry but preserves a second one: entities are born in **two** structural places — the `children` array and entity-bearing component fields (`Room.movement_nodes` etc.). Four consumers must traverse both: asset inference (#563), save/load, `postLoad` hook firing, and gizmo registration. "A walker that traversed one path but not the other" is precisely the bug shape #467 / #470 / `flying-platform-labelle#286` share — so N independent traversals re-introduce the RFC's own target bug, N times.

## Scope

- One walker utility in `labelle-engine/scene/` that yields every entity in a prefab tree, crossing both `children` and entity-bearing component fields.
- All four consumers call it — no consumer hand-rolls traversal.
- **Cycle detection**: a referenced-prefab cycle (`A -> B -> A`) is a load-time error reporting the **full chain**, not just "cycle detected." Applies to both inference (static) and instantiation (runtime).
- Walk order is normative and documented (feeds the hook-order spec).

## Acceptance

- A deliberately cyclic fixture errors with the chain in the message.
- Inference, save/load, postLoad, and gizmo registration share one traversal entry point (verified by grep — no second `children` recursion).

Refs: #560, #561, #563.

apotema:
A single shared entity-tree walker for the unified prefab/scene format — ticket #569, part of RFC #560.

**Stacked on #574** (base `feat/562-overrides-merge-rules`). Review #573 → #574 → this in order; the base retargets as each lands.

## What's in

**`src/jsonc/tree_walker.zig`** — the shared walker. Walks a JSONC `Value` entity tree in a documented pre-order, crossing **both** `children` arrays and prefab references nested inside entity-bearing component fields (the `Room.movement_nodes` pattern). Resolves `prefab` references through a `Resolver` interface (`PrefabCache` satisfies it directly). Detects referenced-prefab **cycles** via an expansion stack and reports the full chain (`A -> B -> A`). Visitors are generic — each node carries its `Origin` (root / child / prefab_child / component_field), depth, owning component, and resolved prefab root. Re-exported as `engine.tree_walker`.

## Consumer migration

- **Migrated:** scene/prefab instantiation. `processEntities` runs a walker pre-pass over every top-level entity as a load-time cycle gate before any entity is created; `spawnPrefabImpl` gates runtime `spawnFromPrefab` the same way. The loader previously had **no** cycle detection — a cyclic graph just hit `IncludeDepthExceeded` with no actionable message; now it logs the chain and returns `error.PrefabCycle`.
- **Deferred (follow-up):** save/load prefab tagging, `postLoad` hook firing, asset inference (#563). These walk the tree inline within `loadEntityInternal`; folding them onto the visitor API is best co-designed once #563 lands a second consumer. The gizmo/save-load *mixins* walk the runtime ECS tree, not the JSONC tree — out of scope.

## Tests

`spec/tree_walker_spec.zig` — 5 zspec specs: `children` crossing (pre-order), component-field crossing, direct + multi-hop cycles (full chain in the message), and an acyclic prefab tree. `zig build spec`: 17/17. `zig build test`: green (verified).

Resolves #569.

### DIFF
--- a/spec/spec_tests.zig
+++ b/spec/spec_tests.zig
@@ -9,11 +9,13 @@ const zspec = @import("zspec");
 pub const unified_format_spec = @import("unified_format_spec.zig");
 pub const override_merge_spec = @import("override_merge_spec.zig");
 pub const registry_scan_spec = @import("registry_scan_spec.zig");
+pub const tree_walker_spec = @import("tree_walker_spec.zig");
 
 // Re-export the spec structs so zspec discovers them.
 pub const UnifiedFormatSpec = unified_format_spec.UnifiedFormatSpec;
 pub const OverrideMergeSpec = override_merge_spec.OverrideMergeSpec;
 pub const RegistryScanSpec = registry_scan_spec.RegistryScanSpec;
+pub const TreeWalkerSpec = tree_walker_spec.TreeWalkerSpec;
 
 test {
     zspec.runAll(@This());
--- a/spec/tree_walker_spec.zig
+++ b/spec/tree_walker_spec.zig
@@ -441,6 +441,50 @@ pub const TreeWalkerSpec = struct {
         }
     };
 
+    // ── scene-declared repetition of a prefab is not a cycle ────
+    //
+    // A reference-chain cycle (A's body references B's body
+    // references A's body) is what `error.PrefabCycle` exists to
+    // catch — NOT a scene that legitimately spawns two instances of
+    // the same prefab in a `children` array. The expansion stack
+    // must release the parent's prefab name before recursing into
+    // the parent's scene-declared `children`; otherwise a sibling /
+    // child reference to the same prefab is falsely flagged.
+
+    pub const @"a scene child re-referencing the parent's prefab" = struct {
+        test "is not a cycle (scene-declared, not reference-chain)" {
+            const a = arena.allocator();
+            const src = Corpus.create(.{});
+
+            var table = PrefabTable.init(a);
+            defer table.deinit();
+            // `leaf` is a plain prefab with no children — two
+            // instances are independent finite trees.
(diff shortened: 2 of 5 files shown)


apotema:

BLOCKER
- None.

MAJOR
- `src/jsonc/tree_walker.zig:248` — Cycle checking walks prefab component fields before considering `overrides`/`null` removals, so valid scenes that replace/remove a cyclic entity-bearing field can be rejected with false `PrefabCycle`.
- `src/jsonc/tree_walker.zig:280` — The walker skips an entire component array unless the first item is entity-like, but runtime spawning scans all items, so cycles in later entity-like entries can be missed.

MINOR
- `src/jsonc/scene_loader.zig:220` — `spawnPrefabImpl` requires `root.components` before running the cycle gate, so cyclic prefab trees without root components bypass the new diagnostic path and just return `null`.
- `spec/tree_walker_spec.zig:208` — Specs cover added override traversal only superficially; they do not test override replacement/removal semantics that are now critical to avoiding false cycle errors.

NIT
- `spec/tree_walker_spec.zig:159` — `emptyResolver` returns a resolver containing a pointer to a stack local; the callback ignores it, but it is still a fragile pattern for a reusable test helper.

Overall verdict: the shared walker is a solid direction and the cycle-chain diagnostics are useful, but I would not merge as-is because the traversal is not aligned with the effective component tree after RFC #562 override merging. In particular, walking prefab component fields independently from overrides can block valid scenes, while the first-element array heuristic can miss entities the loader still spawns. Fix those semantic mismatches and add regression specs around override removal/replacement before landing.

PR #575, opened 2026-05-21
Audit scenes/+prefabs/ name collisions (#570)

apotema:
Audits the example projects' `scenes/` and `prefabs/` `.jsonc` files for effective-name collisions, ahead of enabling the unified registry's hard collision error (#561) — ticket #570, part of RFC #560.

## Findings

`RFC-COLLISION-AUDIT.md` (new doc). 94 `.jsonc` files audited:
- **flying-platform-labelle** (90 files) — every effective name unique; all 36 scenes carry a `"name"` matching their basename, all 54 prefabs resolve by basename.
- **labelle-cli** test projects (4) — clean.

**No collisions** — within `scenes/`, within `prefabs/`, or scene-vs-prefab cross-directory.

## Verdict

Safe to enable the unified registry's hard collision error with #561 — no file needs renaming, no `"name"` field needs changing. This **unblocks the eager filesystem registry scan** (the remaining #561 bullet).

## Note

`bouncing-ball`, `labelle-gui/examples/project_1`, and `bakery-game` use legacy **`.zon`** scene/prefab files, not `.jsonc` — invisible to the registry scan, so no collision risk, but they still need the RFC #560 format migration (tracked under #572).

Resolves #570.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #577, opened 2026-05-21
Eager filesystem registry scan (#561)

apotema:
The remaining piece of #561 — the eager filesystem scan of `scenes/` + `prefabs/` into the flat name-keyed registry. Part of RFC #560.

**Stacked on #573** (base `feat/561-unified-loader`).

## What's in

- `prefab_cache.zig` — `scanRegistry` + `scanDir`: a recursive `.jsonc` walk that parses each file, computes its effective name (`unified_format.effectiveName` — `"name"` field or basename), and populates the `PrefabCache` keyed by effective name. A duplicate effective name across any two scanned files raises `error.DuplicatePrefabName` (same contract as `addEmbeddedPrefab`).
- `scene_loader.zig` — wired into the desktop `loadScene` path. The `scenes/` directory is derived by convention as the sibling of `prefab_dir` (`<dirname(prefab_dir)>/scenes`).

This makes name-divergent files resolvable on desktop and catches scene/prefab cross-collisions at load time — the #570 audit (PR #575) confirmed no real collisions exist, so enabling the hard error is safe.

## WASM / mobile safety

`scanRegistry` returns immediately on `emscripten` (no filesystem — prefabs/scenes arrive via `addEmbeddedPrefab` / embedded sources). Only the desktop `loadScene` path calls it; `loadSceneFromSource` and `addEmbeddedPrefab` are untouched. A missing `scenes/` or `prefabs/` directory is skipped silently.

## Tests

`spec/registry_scan_spec.zig` — 3 zspec specs: prefab found by a divergent `"name"`, scene found via recursive subdirectory scan, collision raising the error. `zig build spec` 9/9; `zig build test` green.

Note for future test authors: the scan now runs on every desktop `loadScene`, so a new test with two same-effective-name files under `prefabs/`/`scenes/` will hard-fail by design.

Completes #561.

### DIFF
--- a/spec/registry_scan_spec.zig
+++ b/spec/registry_scan_spec.zig
@@ -210,4 +210,61 @@ pub const RegistryScanSpec = struct {
             try std.testing.expectError(error.DuplicatePrefabName, result);
         }
     };
+
+    // ── idempotent scan across a scene reload ───────────────────────
+
+    pub const @"loadScene called twice on the same game" = struct {
+        var tmp_dir: std.testing.TmpDir = undefined;
+        var game: Game = undefined;
+
+        test "tests:before" {
+            tmp_dir = std.testing.tmpDir(.{});
+            try tmp_dir.dir.createDir(std.testing.io, "prefabs", .default_dir);
+            try tmp_dir.dir.createDir(std.testing.io, "scenes", .default_dir);
+
+            try tmp_dir.dir.writeFile(std.testing.io, .{
+                .sub_path = "prefabs/widget.jsonc",
+                .data =
+                \\{ "name": "super_widget",
+                \\  "root": { "components": { "Marker": { "id": 99 } } } }
+                ,
+            });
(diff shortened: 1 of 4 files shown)


apotema:

BLOCKER
- None.

MAJOR
- `src/jsonc/scene_loader.zig:78` — `loadScene` rescans into the persistent cache on every call, so a second scene load/reload will treat already-registered files as `DuplicatePrefabName`.

MINOR
- `src/jsonc/prefab_cache.zig:154` — deriving `scenes/` with `dirname(prefab_dir)` breaks when `prefab_dir` has a trailing slash, scanning `prefabs/scenes` instead of sibling `scenes`.

NIT
- None.

Overall verdict: The eager scan covers the intended behavior, but it is not idempotent against the game-lifetime cache it reuses. That makes normal reload / scene-switch paths fragile enough that I would not merge until the scan tracks completed directories or tolerates identical previously-scanned entries.

PR #596, opened 2026-05-27
RFC: flatten unified-format wrappers + bundle shape (v2.0 phase 3)

apotema:
Adds \`RFC-FLATTEN-WRAPPERS-AND-BUNDLES.md\` extending [RFC #594](https://github.com/labelle-toolkit/labelle-engine/pull/594) with three more axes targeting the same structural-noise problem.

## Why

After #594 drops \`root:\`, the dominant FP shape is still:

\`\`\`jsonc
{ "prefab": "rabbit", "overrides": { "Position": { "x": 400, "y": 0 } } }
\`\`\`

Five levels of nesting for "place rabbit at (400, 0)". Across the 50-entry colony scene, \`overrides\` + \`Position\` + \`{x, y}\` wrappers + the scene's outer \`{name, children: [...]}\` are most of the bytes.

## What

Four axes, three new:

1. **(#594) Drop `root:` wrapper.** Already designed.
2. **Drop `overrides:` / `components:` wrappers** via PascalCase/lowercase convention. Component overrides sit as siblings of \`prefab\` on the entry.
3. **File-as-array for bundles.** When the file has no real root entity (scenes; collections of parallel prefabs), the top-level JSON value is an Array of sibling entities. \`children:\` stays valid INSIDE entity objects for the true Parent-of-children case.
4. **\`meta:\` field** for authoring-only side data. Free-form JSON, no propagation, never reaches runtime.

After all four axes, the colony scene reduces to:

\`\`\`jsonc
[
    { "prefab": "ship_carcase", "Position": { "x": 0,   "y": 0 } },
    { "prefab": "ship_carcase", "Position": { "x": 780, "y": 0 } },
    { "prefab": "condenser",    "Position": { "x": 0,   "y": 0 } },
    { "prefab": "kitchen",      "Position": { "x": 156, "y": 93 }, "meta": { "label": "main kitchen" } }
]
\`\`\`

## Bundling

Bundles into v2.0 alongside #594 and engine#592. \`labelle migrate unified\` gains four new transforms (lift \`overrides\`, lift \`components\`, file-object-to-array, drop redundant \`name:\` field) — all byte-offset positional edits with comment preservation, same shape as the existing migrator.

## Open questions

The RFC lists five — most material: does \`name:\` survive as a structural key (for non-default identifier overrides), or drop entirely with folder-prefixed identifiers handling the rare basename-collision case? My read in the RFC: drop \`name:\`, let folder prefix handle collisions, friendly labels go in \`meta:\`.

## Discussion

Not for merging until the open questions land. This is a design doc PR; implementation follows in subsequent PRs (loader extension, audit extension, migrator transforms) against engine#592's v2.0 umbrella.

### DIFF
--- a/RFC-FLATTEN-WRAPPERS-AND-BUNDLES.md
+++ b/RFC-FLATTEN-WRAPPERS-AND-BUNDLES.md
@@ -93,20 +93,36 @@ Three sibling top-level entities; the middle one has its own children-of-it via 
 
 ### Axis 4 — `meta:` field for authoring-only side data
 
-Add a `meta:` structural key for free-form documentation/tooling data that the engine never reads at runtime. Useful for tooltips, editor labels, build-tool tags, debug hints — anything that exists to help humans or tools and isn't needed by gameplay.
+Add a `meta:` structural key for free-form documentation/tooling data that the engine never reads at runtime. Tooltips, editor labels, build-tool tags, debug hints, friendly display names — anything that exists to help humans or tools and isn't needed by gameplay.
+
+Including **friendly display labels via `meta.name`**:
 
 ```jsonc
 [
-    { "prefab": "kitchen", "Position": { "x": 0, "y": 93 }, "meta": { "label": "main kitchen" } },
-    { "prefab": "kitchen", "Position": { "x": 156, "y": 93 }, "meta": { "label": "secondary kitchen" } }
+    { "prefab": "kitchen", "Position": { "x": 0,   "y": 93 }, "meta": { "name": "Main Kitchen" } },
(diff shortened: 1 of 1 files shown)


apotema:
Closing — Gemini flagged two internal contradictions in the RFC that need resolving before this lands:
- **Step 3 vs Step 4**: Step 3 drops `name` for no-root files; Step 4 drops `name` only *unless it diverges*. The migration plan contradicts itself on when `name` is dropped.
- **Bundle meta**: Axis 3 says a bundle is a top-level JSON **array**, but an array can't carry top-level keys — so a bundle file can't have file-level `meta` as written.

Also: entities table is missing the `ref` key (@117). Will revise and reopen (branch kept).

PR #612, opened 2026-06-10
feat(profiler): built-in per-script / per-plugin frame profiler

apotema:
## What
Revives the `profile[]`/`plugin_profile[]` scaffolding that was stubbed during the Zig 0.16 `std.time.Timer` removal, turning it into a working frame profiler.

- New `scene/src/profiler.zig`: monotonic `nowNs()` (clock_gettime via @extern), a cached `recording()` gate on the `LABELLE_PROFILE` env var, a rolling `Stat` (worst/avg), and a worst-first `report()` to the log.
- Un-stubs the timing in `ScriptRunner` (tick/drawGui) and `SystemRegistry` (tick/postTick), feeding the existing per-entry arrays (already exposed via `Game.script_profile_ptr`/`plugin_profile_ptr` for a future overlay).
- `profiling_enabled` flipped from `builtin.mode == .Debug` to always compiled-in, so **ReleaseFast can be profiled** (recording stays gated at runtime; off-path is ~one branch). Lives in the `scene` module so both the engine `ScriptRunner` and the `scene` `SystemRegistry` can reach it; re-exported as `engine.profiler`.

## Use
`LABELLE_PROFILE=1 <game>` → logs `info(profiler)` `[script]/[plugin] <name>: worst=Xms avg=Yms` rankings every ~120 frames. Works headless (stderr).

## Why
Hunting at-scale FPS dips required hand-instrumenting the runners repeatedly; this makes it flip-a-flag and CI-able. Used it to attribute the scheduler and needs_decay dips.

## Test
`zig build` + `zig build test` pass. Companion CLI flag `labelle run --profile` in a separate labelle-cli PR.

Follow-up: an imgui debug-overlay panel (data already exposed for it).

### DIFF
--- a/scene/src/profiler.zig
+++ b/scene/src/profiler.zig
@@ -22,22 +22,55 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 const is_wasm = builtin.cpu.arch == .wasm32 or builtin.os.tag == .emscripten;
+const is_windows = builtin.os.tag == .windows;
 
 const Timespec = extern struct { sec: i64, nsec: i64 };
 extern "c" fn clock_gettime(clk: c_int, tp: *Timespec) c_int;
-extern "c" fn getenv(name: [*:0]const u8) ?[*:0]const u8;
 
 /// Monotonic time in nanoseconds. Desktop only — returns 0 on wasm
 /// (profiling is not wired for the browser build).
 pub fn nowNs() u64 {
-    if (is_wasm) return 0;
-    const clk: c_int = switch (builtin.os.tag) {
-        .macos, .ios, .watchos, .tvos => 6, // _CLOCK_MONOTONIC
-        else => 1, // CLOCK_MONOTONIC (linux et al.)
-    };
-    var ts: Timespec = .{ .sec = 0, .nsec = 0 };
-    _ = clock_gettime(clk, &ts);
-    return @as(u64, @intCast(ts.sec)) * std.time.ns_per_s + @as(u64, @intCast(ts.nsec));
+    if (comptime is_wasm) return 0;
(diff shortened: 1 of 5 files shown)


apotema:
Addressed the review comments in `a8e90ee`:

1. **draw_gui per-window reset** (cursor + gemini) — fixed: `dumpProfile` now calls `e.draw_gui.resetWindow()` alongside `e.tick.resetWindow()` (matches the plugin path in `scene/src/system.zig`, which already reset both).
2. **`nowNs()` Windows linker break** (gemini) — fixed: comptime Windows branch via `std.os.windows.ntdll.RtlQueryPerformanceCounter/Frequency` (note: `QueryPerformanceCounter` is not directly exported; the `Rtl*` ntdll forms are). `clock_gettime` is now only referenced under `if (comptime builtin.link_libc)`. Verified: `x86_64-windows` object has 0 `clock_gettime` references.
3. **`recording()` no-libc break** (gemini) — fixed, with a correction: `std.posix.getenv` does **not** exist in Zig 0.16 (`std.os.environ` was removed), so the right fix is to gate `getenv` behind `builtin.link_libc` and keep the profiler off on no-libc. Also caught + fixed the same no-libc issue in `nowNs()` (`clock_gettime` is a libc symbol) — no-libc Linux now uses `std.os.linux.clock_gettime`. Verified via a no-libc `x86_64-linux` cross-compile.
4. **`{d:6.2}` format specifier** (gemini) — verified it compiles and runs fine in Zig 0.16 (a width without explicit alignment is valid); left unchanged.

`zig build` + `zig build test` (28/28) green; cross-compiles clean for Windows (libc + no-libc), Linux no-libc, and native macOS.

Issue #310, opened 2026-02-24
Generated main.zig missing renderStandaloneGizmos and per-frame clearGizmos


apotema:
## Problem

The generated `main.zig` game loop is missing two calls needed for standalone gizmos (from `game.gizmos.drawRect()`, `drawLine()`, etc.) to work:

1. **`game.gizmos.renderStandaloneGizmos()`** — standalone gizmos are never drawn to screen
2. **`game.gizmos.clearGizmos()`** — without per-frame clearing, gizmos accumulate every frame (scripts call `drawRect`/`drawLine` in their `update()`, so thousands of overlapping shapes pile up)

### Current generated loop

```zig
while (game.isRunning()) {
    const dt = game.getDeltaTime();
    scene.update(dt);
    game.getPipeline().sync(game.getRegistry());

    const re = game.getRetainedEngine();
    re.beginFrame();
    re.render();
    game.processPendingScreenshot();
    re.endFrame();
}
```

### Expected loop

```zig
while (game.isRunning()) {
    const dt = game.getDeltaTime();
    game.gizmos.clearGizmos();
    scene.update(dt);
    game.getPipeline().sync(game.getRegistry());

    const re = game.getRetainedEngine();
    re.beginFrame();
    re.render();
    game.gizmos.renderStandaloneGizmos();
    game.processPendingScreenshot();
    re.endFrame();
}
```

### Context

Discovered while integrating `labelle-needs` into bakery-game. The `needs_manager.zig` script draws sleep level bars above workers using `game.gizmos.drawRect()` in its `update()`. Entity-attached gizmos (from `.gizmos` in prefab .zon files) render fine through the retained engine — only standalone gizmos are affected.

The template in `src/templates/` needs updating so `zig build generate` produces the correct loop.

PR #620, opened 2026-06-13
feat(android): route immersive-mode through core seam (#310 stage 2)

apotema:
Stage 2 of #310 — route engine's Android immersive-mode through labelle-core's `AndroidBackendContext` seam (Stage 1, core, already merged). Removes engine's last direct sokol coupling.

## What (single file: `src/android.zig`, +33/-15)
- Removed `extern "c" fn sapp_android_get_native_activity()`.
- `enableImmersiveMode()` now sources the `ANativeActivity*` from the seam instead of the sokol symbol:
  ```zig
  const ctx = core.android_backend.get() orelse return;       // no backend → immersive no-op
  const na: *ANativeActivity = @ptrCast(@alignCast(ctx.get_native_activity() orelse return));
  ```
  Public signature + all the JNI/decor-view hook logic unchanged — only the activity source swapped. Graceful no-op when no context is registered (matches core).
- Imports core as `@import("labelle-core")` (engine's existing convention; core re-exports `android_backend`).

Engine no longer references any `sapp_*`/sokol symbol. Desktop unaffected (immersive is Android-only).

## Verification (Zig 0.16.0)
- `zig build test` → 28/28 pass; `zig build` → exit 0.
- Android compile-check (`zig build-obj -target aarch64-linux-android` against the merged core seam) → exit 0; `nm` shows **no `sapp` symbol** in the object.
- Grep: no `sapp_*` / `extern "c" fn sapp_...` in `src/` (the only remaining `extern "c"` is NDK `__android_log_write`).

## Note for next stages
Immersive (and core's Android gamepad source) now stay inert until a backend adapter calls `core.registerAndroidBackend(...)` at startup. Stage 3 (sokol adapter) registers a context wired to `sapp_android_get_native_activity` + the gamepad JNI glue (keeping sokol-Android green); Stage 4 (bgfx adapter) registers its own + drops the shims.

Engine pins core by relative path and CI clones core `main` (which has the Stage 1 seam), so no version bump is needed here.

### DIFF
--- a/src/android.zig
+++ b/src/android.zig
@@ -18,9 +18,12 @@
 //! The game is a pure `NativeActivity` (`android:hasCode="false"`), so
 //! there is no Java code to call into directly — every framework call
 //! has to go through JNI from C/Zig. We obtain the `ANativeActivity*`
-//! from sokol_app's `sapp_android_get_native_activity()` (bound here as
-//! an `extern "c"` symbol), which exposes a `JavaVM*`, a main-thread
-//! `JNIEnv*`, and the activity `jobject`.
+//! from labelle-core's backend-agnostic Android seam
+//! (`core.android_backend`, labelle-core#310): the active backend
+//! registers an `AndroidBackendContext` whose `get_native_activity`
+//! returns the activity, which exposes a `JavaVM*`, a main-thread
+//! `JNIEnv*`, and the activity `jobject`. The engine therefore links no
+//! backend-specific (`sapp_*`/sokol) symbol of its own.
 //!
 //! We use **two paths**, picked at runtime by API level:
 //!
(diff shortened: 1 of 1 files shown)


apotema:
Bot triage:

- **`if (comptime !is_android) return;` doesn't discard the body → compile error on non-Android** (Gemini, medium) — **declined, false premise.** The concern assumes `core.android_backend` is unavailable off-Android, but `android_backend.zig` is a **plain cross-platform module** in core (not `is_android`-gated) — `get()` returns `?AndroidBackendContext` on every target. So even when the body is analyzed on desktop it compiles fine; CI's host `zig build test` (28/28) + `zig build` prove it. The `if (comptime !is_android) return;` guard is also pre-existing (this PR only swapped the activity *source* from the sokol extern to the seam, not the guard). No change needed.

Issue #551, opened 2026-05-16
preview_mode: Windows port — replace POSIX TCP/shm with portable primitives


apotema:
## Context

PR #547 (engine v1.37.3) ships the macOS IOSurface producer. PR #544 (engine v1.37.0) ships the SHM ring producer. Both APIs are POSIX-only: the TCP control channel in \`preview_mode.zig\` uses \`read\`/\`write\` against an \`fd: c_int\`, and \`preview_shm.zig\` uses \`shm_open\` + \`mmap\` via \`std.c.O.RDWR\`/\`MAP\` constants that are \`void\` on Windows.

Surfaced as a real blocker by labelle-assembler#126 (sokol D3D11 producer — PR #127): the assembler-side D3D11 codegen is portable and AST-clean for \`x86_64-windows\`, but linking a generated game against the engine fails because the engine's preview control plane refuses to compile on Windows.

Same blocker applies to the GLCORE sokol path on Windows (#124 / slice 1 of assembler#122) — \`os.tag == .linux\` gates it out for now but a future GLCORE-on-Windows config would face the same wall.

## Scope

Port the preview transport so labelle-engine compiles + runs on Windows:

1. **TCP control channel** (\`preview_mode.zig\`): replace \`fd: c_int\` with \`std.Io.net.Stream\` (or whatever 0.16 lands on). The reads/writes are JSON newline-framed today — port to \`std.Io.Reader\`/\`Writer\` and let the stdlib handle the cross-platform syscall mapping. \`std.posix.fcntl\` for non-blocking goes via \`std.Io.net.Stream\`'s nonblocking-mode hooks or a per-platform helper.
2. **SHM ring** (\`preview_shm.zig\`): replace \`shm_open\` + \`mmap\` with Windows' \`CreateFileMappingW\` + \`MapViewOfFile\`. The name space differs (POSIX \`/lbl-prv-...\` vs Windows \`Global\\\\lbl-prv-...\` or local session) — pick a portable naming scheme. \`shm_unlink\` becomes \`CloseHandle\` + reference counting on the mapping name.
3. **macOS IOSurface** (\`preview_iosurface.zig\`): already macOS-only by design (\`if (!macos) return error.PlatformUnsupported\` everywhere). Don't touch — Windows uses SHM mode via labelle-assembler#126's D3D11 staging path.

## Suggested split

Land in slices so each builds on the previous:

[...]

PR #628, opened 2026-06-16
refactor(preview): split preview_mode.zig (<1000 lines/file)

apotema:
## What

Behavior-preserving split of the 1627-line `src/preview_mode.zig` into focused submodules under `src/preview/`, leaving `preview_mode.zig` as a thin re-export facade.

**No API change. No behavior change. No version bump.**

## New layout (all files < 1000 lines)

| File | Lines | Holds |
|------|------:|-------|
| `src/preview_mode.zig` (facade) | 142 | doc + re-exports of the full public surface |
| `src/preview/protocol.zig` | 201 | wire constants, message-shape types, error sets, `parseArgs` |
| `src/preview/socket.zig` | 192 | platform socket I/O shims (#551 POSIX/Windows: read/write/close/non-blocking toggle, errno, pid) |
| `src/preview/connection.zig` | 480 | the `Preview` struct — fields, lifecycle (`connect`/`deinit`/`sendHello`/`sendHeartbeat`/`tickHeartbeat`/`sendBye`), input-event queue, subscription/handshake accessors, `writeFrame`; thin delegating methods for the producer + telemetry surfaces |
| `src/preview/frame_stream.zig` | 450 | PIE viewport SHM + macOS IOSurface producer logic (`beginFrameStream`/`publishFrame`/`endFrameStream` and IOSurface variants, `allocShmName`, `sendFrameOffer`/`sendFramePublished`) |
| `src/preview/telemetry.zig` | 450 | binary telemetry frame emitters (`emitEntityCreated`/`emitComponentChanged`/`emitNodeEntered`/`emitPinValue`/…), subscription-frame parser + non-blocking inbox drain (`pollSubscription`) |

The per-concern producer/telemetry logic lives in sibling modules as free functions taking `*Preview`; the `Preview` struct keeps its identical public method set as thin delegating wrappers. (`writeFrame`/`pushInputEvent` were made `pub` so the sibling modules can reach them — additive only.)

## Public API unchanged

Every previously-public decl remains reachable with the same name/signature via the facade and the `engine.*` root re-exports: `Preview` (and all its methods), `Preview.PublishError`, `ByeReason`, `BinaryFrameKind`, `SnapshotComponent`, `InputEvent`, `FramePixelFormat`, `FrameOffer`, `FrameHandshakeState`, `PendingResize`, `binary_magic`, `protocol_version`, `heartbeat_interval_ms`, `parseArgs`, `preview_shm`, `preview_iosurface`, and the error sets.

## Tests

`zig build test` passes identically before and after. The preview test binaries (which import the public API via the facade) pass unchanged:
- `preview_mode_test` 39/39
- `preview_handshake_test` 7/7
- `preview_frame_stream_test` 10/10
- `preview_iosurface_test` 15/15
- `preview_capture_test` 4/4

(`build.zig` libc-linkage comment updated to point at the relocated `src/preview/socket.zig`; module-level linkage is unchanged.)


apotema:
**Triage of the Gemini comment(s):** the flagged line(s) are code this PR moved **verbatim** (no logic change) — pre-existing, not introduced by the refactor — so changing them is out of scope for a behavior-preserving split. Also, the `comptime` suggestion is redundant here: Zig already comptime-folds an `if` with a comptime-known condition (`builtin.os.tag`/target checks) and does **not** semantically analyze the dead branch. The proof is this PR — CI is green on **both** ubuntu and windows, which is only possible if each platform correctly skips the other platform`\s branch (the POSIX `std.c.*` calls on windows, and vice-versa). If it were a real compile error, one of those jobs would fail. (Re: the four `socket.zig` POSIX/Windows shims — these were moved verbatim from preview_mode.zig and compile on both CI platforms.)

Issue #630, opened 2026-06-16
Scene teardown: O(N²) unloadCurrentScene + scene_entities not cleared in resetEcsBackend


apotema:
Two pre-existing concerns surfaced by Gemini while reviewing the game.zig mixin split (#629). Both are in code that PR moved **verbatim** (only `self: *Self`→`*Game`), so they predate it and were out of scope for a behavior-preserving refactor — filing here to fix properly against the original logic.

1. **O(N²) scene unload** (`game/scene_runtime_mixin.zig` / former `unloadCurrentScene`): per-entity destruction does linear scans (e.g. `untrackSceneEntity` swap-remove scans `scene_entities` for each popped entity). For scenes with many entities this is quadratic. Consider draining the list without the per-entity back-scan (it already pops from the end), or a different bookkeeping structure.

2. **Stale entity IDs after ECS reset** (`game/world_mixin.zig` / `resetEcsBackend`): `resetEcsBackend` wipes the ECS but does not clear `self.scene_entities`. Today the callers (`setSceneAtomic`, `loadGameState`) clear it first, so it's latent — but a direct `resetEcsBackend` call would leave stale IDs that a later `unloadCurrentScene` would `destroyEntityOnly` as invalid handles. Clearing `scene_entities` inside `resetEcsBackend` would make it self-consistent.

Neither changes #629's behavior (verbatim move); both worth fixing on their own.

apotema:
Already fixed by PR #631 (v1.53.1, merged 2026-06-16) — the issue just never auto-closed. Verified on current origin/main: unloadCurrentScene is O(N) (per-entity teardown marker in scene_runtime_mixin + untrackSceneEntity early-return), scene_entities is cleared in resetEcsBackend (world_mixin), and test/scene_teardown_test.zig covers both. Nothing left to do.

PR #631, opened 2026-06-16
fix(scene): O(N) unload + clear scene_entities on resetEcsBackend (v1.53.1)

apotema:
Fixes the two scene-teardown issues in #630.

## Issue 1 — O(N²) scene unload (faster, behaviour preserved)

`unloadCurrentScene` drains tracked entities with
`while (self.scene_entities.pop()) |e| self.destroyEntityOnly(e);`.
`destroyEntityOnly` calls `untrackSceneEntity`, which does an O(N)
swap-remove **scan** of `scene_entities` to find the entity — but the
entity was already removed by `pop()`, so that scan walked the whole
remaining list every call and found nothing. N destroys × O(N) scan =
**O(N²)** on every scene unload.

**Fix:** add a `tearing_down_scene` guard on `Game`, modelled on the
atomic path (`setSceneAtomic`), which already avoids this by clearing the
tracking lists up front so the per-entity untrack is a no-op. The drain
loop sets the flag (restored via `defer`, panic-safe), and
`untrackSceneEntity` early-returns while it's set.

**Behaviour is unchanged — just faster.** Every tracked entity is still
`pop()`-ed and passed to `destroyEntityOnly` exactly once, so the
`entity_destroyed` hooks (and the `engine__entity_destroyed` dual-emit)
fire identically and no entity is left un-destroyed. The only thing
skipped is a redundant scan that was guaranteed to find nothing. The
guard is scoped strictly to the drain loop, so manual `destroyEntityOnly`
calls outside an unload still untrack via the normal swap-remove path
(covered by a test).

## Issue 2 — `scene_entities` not cleared in `resetEcsBackend` (self-consistent)

`resetEcsBackend` wiped the ECS but did **not** clear `self.scene_entities`
(nor the active-scene entity list). Existing callers (`setSceneAtomic`,
`loadGameState`) clear these up front, so the bug was latent — but a
**direct** `resetEcsBackend` would leave stale, now-dangling IDs that a
later `unloadCurrentScene` would feed to `destroyEntityOnly` as invalid
handles.

**Fix:** `resetEcsBackend` now ends with
`self.scene_entities.clearRetainingCapacity()` and
`self.clearActiveSceneEntities()`, mirroring exactly what the atomic path
clears. Idempotent w.r.t. the callers that already clear up front
(clearing an empty list is a no-op).

## Tests

New `test/scene_teardown_test.zig` (wired into `build.zig`'s `test_files`,
using the in-tree `GameWith(Hooks)` like `pause_hook_test.zig`):

- `unloadCurrentScene` destroys all tracked entities, fires
  `entity_destroyed` **exactly once per entity**, empties `scene_entities`,
  and restores the `tearing_down_scene` guard.
- `unloadCurrentScene` on an empty scene is a no-op.
- The guard is drain-scoped: a manual `destroyEntityOnly` still untracks
  via the swap-remove path.
- `resetEcsBackend` clears `scene_entities` (the new behaviour).
- `resetEcsBackend` is idempotent on an already-empty list.

## Verify

- Baseline `zig build test`: 557/557 passed.
- After: 562/562 passed (+5 new tests), `zig build` clean.
- No public API changed; version bumped 1.53.0 → 1.53.1.

Refs #630

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -149,6 +149,9 @@ pub fn build(b: *std.Build) void {
         // Runtime Scheduler for flow `Delay` nodes (#48 / #25 Stage 2).
         // Pause-aware timer wheel reusing the gameplay clock.
         "test/scheduler_test.zig",
+        // #630 — scene-teardown: O(N) `unloadCurrentScene` drain (guarded
+        // untrack) + `resetEcsBackend` clears `scene_entities`.
+        "test/scene_teardown_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,7 +1,7 @@
 .{
     .fingerprint = 0xe8a81a8dc7f48c87,
     .name = .engine,
-    .version = "1.53.0",
+    .version = "1.53.1",
     .minimum_zig_version = "0.16.0",
     .dependencies = .{
         .labelle_core = .{
--- a/src/game.zig
+++ b/src/game.zig
@@ -312,13 +312,16 @@ pub fn GameConfig(
         /// for every entity they create.
         scene_entities: std.ArrayList(Entity) = .empty,
         /// Set for the duration of a full scene-entity drain
-        /// (`unloadCurrentScene`). While true, `untrackSceneEntity`
-        /// skips its O(N) swap-remove scan: the drain already owns the
-        /// list and removes each entry as it goes, so the per-entity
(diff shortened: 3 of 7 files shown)


apotema:
The HIGH findings (Gemini + Cursor: global `tearing_down_scene` flag → sibling double-destroy) are **resolved** in commit 8d78e79 — replaced with `current_teardown_entity: ?Entity` (skip untrack only for the entity being drained; siblings still untrack), exactly as suggested, plus a test locking it. CI green.

PR #632, opened 2026-06-16
feat(assets): carry GPU-compressed blobs through the catalog decode→upload

apotema:
## What

The async streaming asset catalog (engine#450) loads atlases via an injected `ImageBackend` that strictly splits worker-thread `decode` (CPU/stb) → main-thread `upload` (RGBA `uploadTexture`). This path **bypasses gfx's synchronous `loadTextureFromMemory` compressed seam**, so GPU-compressed (ASTC) blobs always fall through to `stb_image` and fail with `error.LoadFailed`.

This PR is the **engine half**: let `DecodedImage` carry a raw compressed blob from `decode` → `upload` so the adapter can route it to the backend's compressed-upload path.

## Changes

- Add `compressed: bool = false` to `DecodedImage` in `src/assets/loaders/image.zig`. When `true`, `pixels` holds the raw ASTC blob (not RGBA8) — the worker `decode` does no CPU work, it just dupes the blob and reads dims from the header.
- Add the same flag to the internal `DecodedPayload.image` struct in `src/assets/loader.zig`.
- Thread the flag through the loader's `decode` (return) and `upload` (the `b.upload(...)` call).

Default is `false`, so the RGBA path is completely unchanged.

## Scope / testing

The `ImageBackend` adapter is supplied by the assembler at codegen time (not in this repo), so this PR alone does **not** exercise the compressed branch end-to-end — that's expected. Paired **gfx** and **assembler** PRs complete the path.

- `zig build` — clean.
- `zig build test` — all pass, including the image-loader mock-backend `decode→upload→free` round-trip.

## Context

Part of the ASTC cold-start epic labelle-gfx#269. Found while validating ASTC on the real flying-platform-labelle game, where the menu atlas returned `error.LoadFailed` until this seam was threaded through.

### DIFF
--- a/src/assets/loader.zig
+++ b/src/assets/loader.zig
@@ -53,9 +53,11 @@ pub const Texture = u32;
 /// stubbed until those PRs ship.
 pub const DecodedPayload = union(LoaderKind) {
     image: struct {
-        pixels: []u8, // RGBA8, allocator-owned
+        pixels: []u8, // RGBA8, allocator-owned — OR a raw GPU-compressed
+        // (ASTC) blob when `compressed` is set (uploaded verbatim, no decode).
         width: u32,
         height: u32,
+        compressed: bool = false,
     },
     audio: struct {
         samples: []i16, // interleaved PCM, allocator-owned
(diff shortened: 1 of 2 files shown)


apotema:
Added unit test coverage for the compressed decode→upload path (review finding, MEDIUM): `test "image loader: compressed decode→upload round-trip"` in `src/assets/loaders/image.zig`. The mock backend now carries a `decode_compressed` flag; the test sets it, runs decode→upload through the loader, and asserts the `compressed` bit survives into the DecodedPayload and is observed by the backend's `upload` (`last_uploaded.?.compressed`), with the CPU blob freed on success. `zig build` + `zig build test` green (564/564).

PR #638, opened 2026-06-19
RFC: project-configurable Y-axis convention (gfx#274)

apotema:
RFC for **labelle-gfx#274** — the implicit, inconsistent Y-axis split: entity positions are y-up (renderer flips), `screenToDesign` is y-down (raw mouse), and the `.line` Shape adds `end` to the already-flipped position (so `end.y` must be hand-negated).

## Proposal
A top-level **`.y_axis = .up | .down`** in `project.labelle`, defaulting to the screen-native **`.down`** (the renderer's internal space is already y-down); existing games opt into `.up` with a one-line declaration that is a pure no-op under Q1→(b). `.y_axis` defines the **logical** convention; the renderer is the only layer that knows screen/NDC space and applies/skips the flip accordingly. `screenToDesign` stays raw — the axis-aware picking surface is the new `screenToLogical`. The `.line` `end` fix (compose `position + end` in logical space, flip the final endpoint once) falls out for both settings.

## Status
The design has **landed** (core `YAxis` + canonical transform, gfx flip routing, engine `screenToLogical`/`yAxis()`); rollout is tracked in the epic labelle-engine#640. This PR ratifies the design record — revision 6 reconciles the prose with the implementation per review. Renders the doc at `RFC-Y-AXIS-CONVENTION.md`.

### DIFF
--- a/RFC-Y-AXIS-CONVENTION.md
+++ b/RFC-Y-AXIS-CONVENTION.md
@@ -1,6 +1,6 @@
 # RFC: Project-configurable Y-axis convention
 
-**Status:** Draft (revision 4 — adds Q6 sprite-pivot direction, Q7 fix-independence, + a pre-migration scene audit)
+**Status:** Draft (revision 5 — scene audit done: FP + ricochet both y-up; Q2 promoted to load-bearing)
 
 **Tracking:** labelle-gfx#274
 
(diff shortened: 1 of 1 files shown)


apotema:
## Sakana review

**Verdict: Strong design, but please reconcile the RFC status/prose before ratifying.**

The Y-axis design is coherent and the key claims check out:
- The `.line` bug is real: the rendered line adds `line.end` after the position has already been flipped.
- `labelle-core` is the right home for `YAxis` and the canonical transforms.
- Keeping `screenToDesign` raw while adding `screenToLogical` is the right migration choice for existing `.up` games.
- Promoting the camera path to load-bearing is correct: flying-platform is camera-based, so `.up` must reproduce today’s camera behavior exactly.

### Findings

**Medium — the Goals section still says `screenToDesign` agrees with logical coordinates.**

The accepted direction later in the RFC is Q1 → (b): keep `screenToDesign` raw/y-down and add `screenToLogical` for axis-aware coordinates. Under `.up`, `Position` is y-up while `screenToDesign` remains y-down, so the agreeing surface is `screenToLogical`, not `screenToDesign`.

Please update the goal/table wording accordingly. In particular, the proposal table row for `.up` still says `screenToDesign` returns “y-up or raw — see Q1”; Q1 is effectively resolved by the migration plan and should be reflected earlier.

**Medium — RFC status says Draft/not implementing, but the design appears already implemented on main.**

The core/engine/gfx pieces for `YAxis`, `toScreenY`, `screenToLogicalY`, `screenToLogical`, `yAxis()`, and gfx `toScreenY` routing are already present on current main branches. The RFC should either be promoted to Accepted/Implemented or explicitly say it is documenting/ratifying work that has already landed.

**Low — PR description is stale.**

The PR body still says default `.up = today's behavior`, but the RFC now defaults to `.down` and requires existing games to opt into `.up` during the transition.

### Recommendation

Resolve the wording/status drift, then accept. The architecture is sound; the remaining risk is implementation validation around camera composition and sprite/shape offset audits, not the core direction.

— Sakana

Issue #651, opened 2026-06-30
Epic (umbrella): Packs — LLM-friendly feature authoring (initiative tracker)


apotema:
**Umbrella / initiative tracker** for **Packs** — making labelle easy for an LLM (or human) to add a game feature.

## North star
> A small change, or a new feature, should require understanding **one realm**, not the whole system.

A *realm* is a bounded slice of functionality with a clear interface — a **pack**. "Files × footguns to add a feature" is the operational proxy; realm boundaries are the cause.

## What Packs is
A convention that composes labelle's existing primitives (events / prefabs / scripts / components) into self-contained, additive feature bundles, at three tiers — **feature → local pack → published pack** — with a single promotion path (promote by `mv`, not rewrite). Communication: **events up** (notify, decoupled), **reads + writes down** (published **queries** + **commands**), strict cross-pack **encapsulation enforced by the module graph** (a bypass is a compile error, not a style nit).

Most of the value is game-side convention; the engine/toolchain work is narrow and verified against source.

## Design
- [x] **Design RFC** (proving ground, Flying-Platform): Flying-Platform/flying-platform-labelle#561 — full convention: 3 tiers, communication model (events up / reads+writes down), enforcement (module isolation + two-tier registry partition), `pack.labelle` schema, `contracts` pack + sparse vocabulary, agent manifest, FP decomposition. **Merged to `main` 2026-07-01** (`docs/RFC-packs.md`).
- **Engine + toolchain epic** (parent): labelle-toolkit/labelle-engine#650

## Implementation tickets (PR-sized, by phase)

**Phase 2–3 (EASY, independent — can start anytime):**
- [ ] labelle-toolkit/labelle-engine#653 — `entitiesWith(Tag)` roster primitive (query-cache + borrowed-slice lifetime)
- [ ] labelle-toolkit/labelle-assembler#442 — pack/feature manifest (extend `flow_catalog.json` → agent schema)

**Phase 4 (the structural work — engine #652 gates the rest):**
- [ ] labelle-toolkit/labelle-engine#652 — **visibility-preserving registry unification + two-tier component visibility + per-pack partition** *(the crux/HARD)*
- [ ] labelle-toolkit/labelle-assembler#439 — pack dir-scan (`convention_dirs = copy_and_scan`) *(depends on #652)*
- [ ] labelle-toolkit/labelle-assembler#440 — invisible `<pack>__` namespacing + local→prefixed `.jsonc` rewrite
[...]

Issue #650, opened 2026-06-29
Epic: Packs — engine + toolchain support for LLM-friendly feature authoring (FP RFC #561)


apotema:
**Discussion / epic** for the engine + toolchain support behind **Packs** — a convention that makes adding a game feature a self-contained bundle (events + component + script) instead of a horizontal smear, scaling feature → local pack → published pack.

Design RFC (proving ground, flying-platform-labelle): https://github.com/Flying-Platform/flying-platform-labelle/pull/561

Most of the Packs value is **game-side convention** (labelle's events/prefabs/scripts/serializer already carry it). This epic tracks only what **labelle itself** must implement, all verified against source.

### labelle-engine
- [ ] **Unify the two component-registry APIs.** Game-root components feed `ComponentRegistry(...)`; plugin components feed `ComponentRegistryWithPlugins([plugins])` (`scene/src/component.zig`, codegen `registries.zig:54`). A pack's *directory-scanned* components need to land in the unified set. **This is the architectural crux** (the rest depends on it). ⚠️ **Must preserve per-pack visibility** — unify storage + registration WITHOUT flattening the type namespace so every pack can see every type (see Isolation & enforcement below). _MODERATE._
- [ ] **`entitiesWith(Tag)` roster primitive** — a dirty-tracked cached collection on `Game` (~50 LOC in `game.zig`; today every `view()`/`collectEntities` is a fresh scan). Backs pack query-surface caching + the event-sourced aggregate. _EASY, independent._

### labelle-assembler
- [ ] **Finish `convention_dirs = copy_and_scan`** so a pack's `components/ events/ prefabs/ hooks/` are scanned like the game root. The machinery is half-built: `plugin.labelle` already has the `convention_dirs` modes (`plugin_manifest.zig:74`) and `scripts/` is already dir-scanned (`script_scanner.scanPluginDir`, `root.zig:634`), but for the other dirs the scanned names are **computed and discarded** today. _MODERATE; depends on engine #1._
- [ ] **Namespacing + collision rule — invisible `<pack>__` auto-prefix.** Manifest-name-derived `<pack>__<Name>` prefix (mirrors the existing `<plugin>__event`, `plugin_registries.zig:162`) applied to pack components/prefabs/pack-local events **in the generated global registry only** — authors use local names in their realm. Two parts:
[...]

PR #654, opened 2026-06-30
packs: entitiesWith(Tag) roster primitive (query-cache + borrowed-slice lifetime)

### LINKED ISSUE #653: packs: entitiesWith(Tag) roster primitive (query-cache + borrowed-slice lifetime)
**`entitiesWith(Tag)` roster primitive (Phase 2 · EASY · independent).**

A dirty-tracked cached collection on `Game` (~50 LOC in `game.zig`) — today every `view()`/`collectEntities` is a fresh scan. Increment a generation counter on component add/remove; lazy re-query on read when stale; expose `pub fn entitiesWith(self, comptime includes) []const Entity`.

Backs: pack **query-surface caching** (`citizens.idleWorkers()` borrows from this cache) and the **query lifetime contract** — list-returning queries hand back a slice **borrowed from the roster cache** (engine-owned, valid until the next structural mutation), not a caller-owned `collectEntities` alloc.

---
Part of **Packs** — RFC Flying-Platform/flying-platform-labelle#561 · umbrella labelle-toolkit/labelle-engine#651 · epic labelle-toolkit/labelle-engine#650.

apotema:
## What

Implements the `entitiesWith(Tag)` roster primitive from the **Packs** initiative — a dirty-tracked, **cached** collection on `Game` (`src/game.zig`). Today every `view()` / `collectEntities` is a fresh ECS scan; this adds a lazily-refreshed cache so repeated reads of the same tag-set within a mutation-free window are free.

### API

```zig
pub fn entitiesWith(self: *Self, comptime includes: anytype) []const Entity
```

Returns a slice **borrowed from an engine-owned cache**, valid only until the next structural mutation. The lifetime contract: callers must **not** free it, and must **copy** it if they need to retain it past a mutation. This backs pack query-surface caching (e.g. `citizens.idleWorkers()` borrows from here) and the borrowed-slice lifetime contract for list-returning queries.

### How it works

- **Epoch counter** (`roster_generation`): a monotonic `u64` bumped (`+%=`) on every structural mutation — `addComponent` / `removeComponent` (genuine adds only via `setComponent`'s `is_update` guard, so per-field updates don't thrash the cache), `createEntity` / `destroyEntity` / `destroyEntityOnly`, and `setActiveWorld` / `resetEcsBackend` (world switch + ECS rebuild).
- **Lazy re-query**: a read compares its cached slot's stored `gen` against the current epoch; on a miss it re-scans the ECS view into the slot and stamps the epoch.
- **Cache structure** (v1 design choice): a small fixed map of **8 slots** keyed by a comptime hash (Wyhash over the component type names) of the tag-set. Distinct tag-sets cache independently; beyond 8 the least-recently-keyed victim is evicted key-indexed and refilled on demand. Single-cache / small-fixed-map is explicitly acceptable for v1 per the issue — caching every distinct tag-set with zero eviction is a follow-up.
- Slot lists are engine-owned (`std.ArrayList(Entity)`) and freed in `Game.deinit` (leak-free under `testing.allocator`).

### Documented limitations / deferred

- **`setPosition` does not bump the epoch.** It is a per-frame value mutation (Position is essentially always present at entity creation), so bumping there would defeat the cache. First-adding `Position` via the generic `addComponent` / `setComponent` path *does* bump correctly. Documented inline.
- Direct `game.ecs_backend.addComponent(...)` (the escape-hatch used in some existing tests) bypasses the epoch bump — use the `Game`-level methods to keep rosters coherent. Documented inline.
- Comptime hash key has a negligible collision risk; ordering of the include tuple affects sharing (same types, different order ⇒ separate slots, still correct).

## Tests

New `test/entities_with_roster_test.zig` (wired into `build.zig`), 8 cases — all green, no leaks:
- roster returns exactly the tagged entities; multi-component intersection
- repeated reads return the same borrowed slice (same `ptr`)
- cache invalidates on component **add**, component **remove**, and entity **destroy**, reflecting the latest set after refresh
- distinct tag-sets cache in independent slots

`zig build` and `zig build test` both pass (full suite green).

Closes #653

Part of **Packs** — umbrella labelle-toolkit/labelle-engine#651 · epic labelle-toolkit/labelle-engine#650 · RFC Flying-Platform/flying-platform-labelle#561.


apotema:
Rebased on `main` and addressed the review. Green + conflict-free (`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`; ubuntu + windows CI pass; full suite 610 tests).

### Conflict resolution
`main` advanced to v1.64.0 with a large mixin-extraction refactor (plus the font refactor), which collided in `src/game.zig` across 5 hunks (`deinit`, entity management, view/component/world/lifecycle accessors). I kept **main's** mixin structure everywhere and **re-homed the #653 roster additions onto it** rather than reverting main's move:
- `deinit`'s roster-cache free -> `game/lifecycle_mixin.zig`.
- `bumpRoster()` now lives in `game.zig` and is `pub` (mixins call it across file boundaries); the create/destroy, add/set/remove-component, and world switch/reset invalidation calls were re-applied inside `entity_mixin.zig`, `components_mixin.zig`, and `world_mixin.zig`.
- `entitiesWith` / `rosterKey` / `rosterSlotFor` stay in `game.zig`.

### Findings applied
- **Round-robin eviction (HIGH)** — added `roster_evict_cursor` and replaced `key % roster_cache_slots` in `rosterSlotFor` with match -> free/invalid -> evict-at-cursor(+advance). No more conflict-miss thrashing.
- **Order-independent key (MEDIUM)** — `rosterKey` now sorts the component type names (comptime insertion sort) before hashing, so `.{A,B}` and `.{B,A}` share one slot. Added a test asserting identical backing pointer for both orderings.
- **OOM logging (MEDIUM)** — the `append` OOM path in `entitiesWith` now `self.log.err(...)`s (the engine's logger on `Game`) before returning the best-effort partial roster.
- **Roster staleness on direct `ecs_backend` mutations (CodeRabbit/Codex, Major)** — closed the gaps: `setParent`/`removeParent`, `tagAsPrefabInstance`/`tagPrefabChildren`, visual `add/remove` (Sprite/Shape/Text/Icon/Gizmo), video `add/remove`, and `setPosition` (guarded to only bump on first-time Position add so per-frame moves don't thrash) all bump now. Save/load's rebuild routes through `resetEcsBackend`, which bumps.

### Skipped (with reason)
- **`@panic` on roster OOM (CodeRabbit/Codex suggestion)** — kept the log-and-return-best-effort behavior instead. `entitiesWith` is intentionally non-fallible with a documented borrowed-slice contract; the codebase's convention elsewhere (e.g. event emit) is log-and-continue, and panicking a game on a transient roster-refill OOM is more aggressive than warranted. The added `log.err` makes a truncated frame diagnosable.
- **Borrowed-slice invalidated by eviction within one tick (Codex P2)** — this is the documented v1 lifetime limitation (a borrowed roster is valid only until the next structural mutation *or* eviction pressure across >8 distinct tag-sets); callers needing to hold one roster while querying others should copy. Noted as a known v1 constraint, not a merge blocker.

### Tests
`zig build` + `zig build test` both pass. The original 8 roster tests still pass; added a 9th for the order-independent key.

Issue #652, opened 2026-06-30
packs: visibility-preserving registry unification + two-tier component visibility (.global/.pack) + per-pack partition


apotema:
**The crux of pack isolation (Phase 4 · HARD · gates the assembler dir-scan).**

(a) **Unify the two component-registry APIs** — game-root `ComponentRegistry(...)` vs plugin `ComponentRegistryWithPlugins([...])` (`scene/src/component.zig`; codegen `registries.zig:54`) — so a pack's dir-scanned components land in one set.
(b) **Per-component visibility** `.visibility = .global | .pack` (default `.pack`/private). `.global` = shared registry, any pack may name (contracts facets like `Locked`, engine primitives); `.pack` = private to the owner.
(c) **Per-pack registry partition** — each pack's resolvable registry = `global ++ its own private`. Foreign-private names then **miss at comptime**, closing the string-escape hole (`getType("citizens__Worker")` / `entityHasNamed(...)` — existing plugin scripts already use these).

**Constraints:** storage stays **one shared ECS backend** (the registry is only a name→type VIEW); MUST unify *storage/registration* without flattening *visibility*. Cheap because queries return opaque handles, so a pack never needs to name a foreign component anyway.

---
Part of **Packs** — RFC Flying-Platform/flying-platform-labelle#561 · umbrella labelle-toolkit/labelle-engine#651 · epic labelle-toolkit/labelle-engine#650.

apotema:
Engine-side slice merged in PR #655 (per-component `.visibility` + `ComponentView`/`PackView` partition; foreign-private names are a compile error). The remaining registry *unification* (single dir-scanned component set + `PackView` wired into generated `Game`/dispatch) is assembler codegen, tracked in labelle-toolkit/labelle-assembler#439.

PR #655, opened 2026-06-30
feat(packs): two-tier component visibility + per-pack registry partition (#652)

apotema:
Part of #651, #652. Implements the **name-resolution core** of the Packs isolation model (RFC Flying-Platform/flying-platform-labelle#561 §6 "Enforcement" / two-tier visibility).

### 1. Per-component visibility (`scene/src/component.zig`)
A component declares its tier with a single `pub const`, mirroring how save behavior is declared via `save_policy`/`Saveable`:

```zig
pub const Locked = struct {
    pub const visibility = .global;   // any pack may name this component
    by: u64 = 0,
};
pub const Worker = struct {           // no decl → defaults to `.pack` (private)
    hunger: f32 = 0,
};
```

- `Visibility = enum { global, pack }`
- `getVisibility(T) Visibility` — defaults to `.pack`; supports both the enum-literal form and the explicitly-typed `pub const visibility: Visibility = .global`.
- `isGlobal(T)` convenience.

### 2. Per-pack registry partition (the escape-hole closure)
- `ComponentView(FullRegistry, allowed_names)` — a restricted **view** that delegates to any existing `ComponentRegistry` / `ComponentRegistryWithPlugins` but only resolves an allow-list:
  - `has(name)` → true only if allowed **and** known to the full registry
  - `getType(name)` → the type for an allowed name; **compile error** for a disallowed (foreign-private/unknown) name
  - `entityHasNamed(...)` → dispatches through `getType`, so a foreign-private name is a compile error here too
  - `names()` → the visible, defined names
- `PackView(FullRegistry, own_names)` = all `.global` components of the full registry `++` the pack's own private names. Foreign-private names are absent from the allow-list, so a script calling `getType("citizens__Worker")` / `entityHasNamed(ecs, e, "citizens__Worker")` on a foreign-private name now **fails to compile** instead of silently reaching into another pack's data.
- `globalNames(FullRegistry)` enumerates the shared facets.

All surfaced via `engine.Visibility`, `engine.getVisibility`, `engine.isGlobalComponent`, `engine.ComponentView`, `engine.PackView`, `engine.globalComponentNames` (and the `scene` submodule roots).

### Hard constraint honored
Storage stays **one shared ECS backend**. The full/global registry used by the serializer + ECS is **unchanged** — it still resolves every name. The partition is purely a name→type *lens* over the existing comptime map; no component data moves.

## Tests (`test/component_visibility_test.zig`, 12 cases)
- `.global` (enum-literal + typed) and `.pack` (default + explicit) visibility reads.
- `globalNames` lists only the global facets.
- A `PackView` resolves its own private components + all globals, and **misses** foreign-private names (`has`/`isAllowed` → false), while the full registry still resolves them.
- `entityHasNamed` dispatch through the view.
- Empty allow-list view resolves nothing.
- The compile-error path (foreign-private `getType`) is documented in-file and verified out-of-band: `Citizens.getType("Ship")` fails with *"Component 'Ship' is not visible to this pack…"*.

`zig build` and `zig build test` both pass (full suite green; new file = 12/12).

## Deferred (honest scope)
This PR is parts **1 + 2** plus the *groundwork* for part 3 — done as a clean, reviewable slice rather than cramming the whole ticket:

- **Full registry unification (assembler codegen).** `ComponentView`/`PackView` already work uniformly over both `ComponentRegistry` and `ComponentRegistryWithPlugins`, so the *view* layer is unification-ready. But emitting a **single** dir-scanned component set is an `labelle-assembler` codegen change (`registries.zig:54`, a different repo) — out of scope here. Follow-up: have the assembler (a) collect dir-scanned pack components into one `ComponentRegistryWithPlugins`, (b) emit a `PackView` per pack keyed on each pack's own component names, and (c) hand each pack's scripts their `PackView` instead of the global registry.
- **Wiring `PackView` into script dispatch / `Game`.** `Game` still exposes the full `ComponentRegistry` (`ComponentRegistry = ComponentsType`). Routing pack scripts through their partition is the enforcement step that the assembler change above unlocks.
- **Pack-name namespacing convention** (e.g. `citizens__Worker`) is assumed by the escape-hole framing but not minted here; it belongs with the dir-scan emitter.

No behavior changes for existing games: the new symbols are additive and nothing consumes `PackView` yet.


apotema:
Rebased onto `main` (now v1.65.0) and addressed review.

**Merge conflict** — only `build.zig` conflicted (the `test_files` list). Kept both sides: `component_visibility_test.zig` (#652) plus main's `scene_teardown_test.zig` and `y_axis_test.zig`. `build.zig.zon` auto-merged; version already matches main at 1.65.0.

**Review comments (both applied):**
- **coderabbit** (`PackView` incompatible with `ComponentRegistryMulti`): `globalNames`/`PackView` need `FullRegistry.names()`, which `ComponentRegistryMulti` doesn't expose. Added a comptime guard that emits an explicit `@compileError` naming the unsupported registry, rather than letting it fail with an opaque "no member named 'names'" error.
- **gemini** (`&result` lifetime in `PackView`): moved the allow-list computation into a nested `Holder` const so it has static, memoized storage per `(FullRegistry, own_names)`, matching the existing `globalNames` pattern. No more returning the address of a comptime-block local.

**Verification:** `zig build` and `zig build test` both green (all 12 visibility tests included). Re-confirmed the isolation invariant: temporarily uncommenting `CitizensView.getType("Ship")` still fails to compile with `error: Component 'Ship' is not visible to this pack (foreign-private or unknown).`

PR #703, opened 2026-07-07
tilemap: engine Tilemap component + .tmx loader + post-sprite render pass (T2 Phase 2)

apotema:
## What

Phase 2 of the minimal T2 tilemap epic: a real, immutable `Tilemap` component that references an embedded `.tmx` asset by name, decodes it through the **gfx 1.21.0** tilemap API, and renders it as a **POST-SPRITE pass**. The engine owns pass ordering (entities first, tilemaps after); gfx's `RetainedEngine` is untouched.

## Load-bearing design decision (the render-orchestration unknown)

**The engine module takes NO labelle-gfx dependency** — the existing invariant (`game.zig`: "The engine does NOT depend on labelle-gfx — it accesses component types through RenderImpl"). The design brief assumed a direct `gfx.TileMap` / gfx pin on the module; grounding against `origin/main` showed that would break the invariant and the "purely additive" bar.

Instead, the engine reaches gfx's `TileMap` decoder + `TileMapRenderer` **through the renderer plugin** — `RenderImpl.TileMapRendererType`, `loadTextureFromMemory`, `getTextureInfo` — exactly as it reaches `Sprite`/`Shape` and the `drawMesh` seam. All the type reflection (naming `TileMap`, `Tileset`, the backend `Texture`, the resolver) is centralised in `src/tilemap_runtime.zig`. The whole feature is comptime-gated on `tilemap_supported`, so stub renderers get a `void` side table + no-op methods — zero cost, no behaviour change for existing consumers.

**Texture crux:** the engine catalog stores `u32` texture ids, but the gfx resolver seam needs a `Backend.Texture`. Bridged via the SAME path sprites use — `RenderImpl.loadTextureFromMemory` (upload) → `getTextureInfo(id).backend_texture` (resolve) — so tileset images share the sprite texture registry and are unowned by the tilemap renderer (`load_unresolved_from_filesystem = false`).

**Asset lifecycle:** a `LoaderKind.tilemap` catalog variant would touch ~13 files / ~86 exhaustive switch sites AND can't hold a gfx `TileMap` in the gfx-agnostic `DecodedPayload` union — that fails "minimal/additive". Used a dedicated embedded-source registry (`embedded_tilemap_sources` + `addEmbeddedTilemapAsset`), same idiom as `embedded_scene_sources`.

**Save:** `Tilemap.asset_name` is `[]const u8`, which the registry-driven serde can't round-trip — so it lives in the **built-in save channel** alongside `Position`/`PrefabInstance`, persisting only `asset_name`. Load rehydrates the runtime by re-decoding. Tilemap-only entities are swept into the save set; ECS reset frees stale runtimes.

The gfx pin (`build.zig.zon`) is **lazy + test-only**: consumed solely by the tilemap test to drive the real shipped gfx 1.21.0 tilemap sub-package headless.

## Files
- New: `src/tilemap.zig`, `src/tilemap_runtime.zig`, `src/game/tilemap_mixin.zig`, `test/tilemap_test.zig`
- Wiring: `game.zig`, `game_init.zig`, `lifecycle_mixin.zig`, `world_mixin.zig`, `entity_mixin.zig`, `loop_mixin.zig`, `jsonc/component_apply.zig`, `editor_api.zig`, `save_load/{save,load}.zig`, `root.zig`, `build.zig`, `build.zig.zon`

## Tests
`745/745` engine tests pass (`zig build test`), incl. 5 new: seam detection, `.tmx` decode, post-sprite ordering + world offset, save→load round-trip of `asset_name`, digest presence. `zig fmt` clean; ReleaseSafe build clean.

## Not in scope (deferred)
- Camera-transform alignment (T2 runs the pass in screen space, `camera_* = 0`) + Z-interleaving with sprite layers → T3.
- Assembler-side `.tmx` + tileset-image embedding → Phase 4.

Do **not** merge — no engine release/version bump (release choreography is handled post-review).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -87,6 +87,9 @@ pub fn build(b: *std.Build) void {
         "test/save_load_mixin_test.zig",
         "test/jsonc/bridge_leak_test.zig",
         "test/jsonc/nested_lifecycle_test.zig",
+        // C2 — a project-registered component named `Tilemap` must win over
+        // the engine built-in in the scene loader (no silent shadowing).
+        "test/jsonc/tilemap_precedence_test.zig",
         "test/scene_ref_test.zig",
         "test/asset_catalog_test.zig",
         "test/audio_loader_test.zig",
--- a/build.zig
+++ b/build.zig
@@ -294,6 +294,38 @@ pub fn build(b: *std.Build) void {
     const assets_single_threaded = b.addTest(.{ .root_module = assets_single_threaded_module });
     test_step.dependOn(&assets_single_threaded.step);
 
+    // T2 tilemap test — proves the engine `Tilemap` component + `.tmx`
+    // decode + post-sprite render pass against the ACTUALLY-shipped gfx
+    // 1.21.0 API. The engine module takes no gfx dependency; this test
+    // reaches gfx's std-only `tilemap` sub-package (the `TileMap` decoder
+    // + `TileMapRendererWith`) through the lazy `labelle_gfx` pin and
+    // drives it over `core.mock_backend.MockBackend`. `lazyDependency`
+    // returns null on the first uncached run and Zig re-invokes after
+    // fetching, so downstream engine-module consumers never pull gfx.
+    if (b.lazyDependency("labelle_gfx", .{ .target = target, .optimize = optimize })) |gfx_dep| {
+        // gfx pins its `tilemap` sub-package in-tree (`.path = "tilemap"`);
+        // reach that module through gfx's own builder. The sub-package is
+        // std-only (no labelle-core), so there is no core-diamond to unify.
+        const tilemap_module = gfx_dep.builder.dependency("tilemap", .{
+            .target = target,
+            .optimize = optimize,
+        }).module("tilemap");
+        const tilemap_test = b.addTest(.{
+            .root_module = b.createModule(.{
+                .root_source_file = b.path("test/tilemap_test.zig"),
+                .target = target,
+                .optimize = optimize,
+                .imports = &.{
+                    .{ .name = "labelle-core", .module = core_module },
+                    .{ .name = "engine", .module = engine_module },
+                    .{ .name = "scene", .module = scene_module },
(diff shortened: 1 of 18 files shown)


apotema:
Addressed in fdd9832.

**Dangling-pointer finding (lines 88/146):** verified against gfx v1.21.0 — `TileMapRenderer.initWithOptions` resolves tileset textures **eagerly** inside the call (loops `map.tilesets`, stores resolved `Texture` handles in `self.textures`) and does **not** retain the resolver. So `&ctx` was only dereferenced while `ctx` was still alive → not a live bug today. But relying on that made the engine silently depend on gfx's resolution *timing* across the repo boundary (a future switch to lazy/draw-time resolution would be a UAF with no compile error), so I stored the context inline in the heap-stable `Runtime` to be correct-by-construction regardless. Thanks — good catch on the latent coupling.

**Lowercase token (line 187):** applied `std.ascii.toLower` so a `.PNG` tileset resolves the same decoder as `.png`.

apotema:
Disposition of the two CodeRabbit findings (both on fdd9832, now on head 00fbf9d):
- **`tilemap_mixin.zig:84` — clear tilemap runtimes on world swaps:** same concern as codex's per-world finding. Handled in 00fbf9d — `renderTilemaps` now reaps side-table entries whose entity isn't live+tilemap-bearing in the *currently active* ECS (so a shelved world's runtimes don't draw), `resetEcsBackend` clears the table via `clearTilemaps` on scene swap / load, and proper per-world scoping (own on `World`, clean in `destroyWorld`) is filed as **#704**. The residual colliding-id edge is multi-world-only and tracked there.
- **`root.zig:66` — re-export tilemap_runtime_mod:** `Runtime`/`supported`/`ImageProvider` are internal plumbing the engine drives itself; games use the `Tilemap` component + `addTilemap` (both re-exported). Leaving the module unexported to avoid widening the public surface for an internal mechanism; will revisit if Phase 4 (assembler) needs it.

PR #707, opened 2026-07-07
fix(tilemap): supported() rejects backends with a non-concrete texture seam — null/headless compiles (v1.75.1)

apotema:
## The regression

engine v1.75.0 + gfx 1.21.0 fails to **compile** the T2 tilemap runtime for any backend built on `GfxRendererWith` — first surfaced by the **null backend** in labelle-assembler's Examples integration test (asset-streaming-smoke), blocking any headless/null consumer:

```
src/tilemap_runtime.zig:72:87: error: unable to unwrap null
  const GetInfoRet = @typeInfo(@TypeOf(RenderImpl.getTextureInfo)).@"fn".return_type.?;
src/game.zig:238:36: note: called at comptime here
  tilemap_runtime.Runtime(RenderImpl)
```

## Root cause

`supported()` gated the feature on decl **name** presence (`@hasDecl`) only. When true, `game.zig` instantiates `Runtime(RenderImpl)`, whose comptime reflection unwrapped `@TypeOf(RenderImpl.getTextureInfo).@"fn".return_type`.

gfx 1.21.0's `GfxRendererWith.getTextureInfo` is declared

```zig
pub fn getTextureInfo(self: *const Self, id: TextureId) ?@TypeOf(self.inner).TextureInfo { ... }
```

Its return type references the `self` **parameter**, which makes the wrapper a **generic** function — so `@"fn".return_type` reflects as `null` and `.?` fails at comptime. This is true for **every** `GfxRendererWith` backend (raylib/sokol/bgfx/null); games on gfx < 1.21.0 escaped only because they lacked the `TileMapRendererType` decl (`supported()` false → `void` no-op). The engine's own tilemap test passed because its `MockRender.getTextureInfo` used a *concrete* return type — not faithful to `GfxRendererWith`.

## Fix

- **`Runtime`** derives its backend `Texture` from the **concrete** `TileMapRendererType.TextureResolver.resolveFn` signature (`fn (…) ?Texture`) instead of the generic `getTextureInfo` return type. `getTextureInfo` is still *called* at runtime, where its generic return type resolves fine.
- **`supported()`** gains `hasReflectableSeam()`: beyond the four `@hasDecl` names it verifies (guarded, no compile error on malformed input) that the seam is concretely reflectable in the exact shape `Runtime` needs — struct `TileMapRendererType` with a `map` pointer + `tilesets`, a `TextureResolver` with a concrete `resolveFn`, and a non-generic error-union `loadTextureFromMemory`. Anything short of that degrades to the `void` side-table no-op rather than failing the build.
- Re-export `engine.tilemapSupported` so a renderer can be probed without configuring a whole `Game`.

Real backends and the null/headless backend now all build with tilemaps functional.

## Tests

- The tilemap test's `MockRender.getTextureInfo` is now **generic**, byte-for-byte the `GfxRendererWith` shape — so the entire existing suite (decode / post-sprite draw / save+load / teardown) exercises the previously-broken seam. Pre-fix the whole binary fails to compile; post-fix 12/12 green.
- `regression: a GENERIC getTextureInfo seam is still supported (null-backend / gfx 1.21.0)` — asserts the seam is generic, then `supported()` true + `TilemapRuntimeType != void`.
- `supported() rejects a backend whose texture seam is not concretely reflectable` — non-reflectable + bare stub → false.
- Verified end-to-end against the **real** null backend: `labelle-assembler` asset-streaming-smoke (`GfxRendererWith(null, …, .up)` + gfx 1.21.0) crashes with the exact CI error against v1.75.0, and generates → builds → runs headless clean against this branch.

`zig build test` green on 0.16.0; `zig fmt` clean; version bumped to **1.75.1**.

Do not merge — release + assembler re-pin gated by the author.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,7 +1,7 @@
 .{
     .fingerprint = 0xe8a81a8dc7f48c87,
     .name = .engine,
-    .version = "1.75.0",
+    .version = "1.75.1",
     .minimum_zig_version = "0.16.0",
     .dependencies = .{
         // Requires labelle-core >= v1.20.0 — font_types aliases core's
--- a/src/root.zig
+++ b/src/root.zig
@@ -63,6 +63,11 @@ pub const Game = game_mod.Game;
 /// asset by name. Reachable on a configured game as `Game.TilemapComp`.
 /// See `src/tilemap.zig` / `src/tilemap_runtime.zig`.
 pub const Tilemap = @import("tilemap.zig").Tilemap;
+/// True when a renderer plugin exposes a CONCRETELY reflectable gfx tilemap
+/// seam. Gates `Game.tilemap_supported` / `Game.TilemapRuntimeType`; exposed
+/// so consumers (and tests) can probe a renderer without configuring a whole
+/// `Game`. See `src/tilemap_runtime.zig`.
+pub const tilemapSupported = @import("tilemap_runtime.zig").supported;
 
 // ── Input ──
 pub const InputInterface = input_mod.InputInterface;
--- a/src/tilemap_runtime.zig
+++ b/src/tilemap_runtime.zig
@@ -87,7 +87,9 @@ fn hasReflectableSeam(comptime RenderImpl: type) bool {
 
     // The shared upload path must expose a concrete (non-generic) error-union
     // return so `Runtime` can name the texture-id handle it threads around.
-    const load_ret = @typeInfo(@TypeOf(RenderImpl.loadTextureFromMemory)).@"fn".return_type orelse return false;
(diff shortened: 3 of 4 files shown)


apotema:
Good catch — fixed in b5a178b. The `loadTextureFromMemory` branch now checks `@typeInfo(...) == .@"fn"` before touching `.@"fn".return_type`, matching the guarded pattern used for `resolveFn` above it, so a backend declaring `loadTextureFromMemory` as a non-function decl degrades to the void no-op instead of compile-erroring — consistent with hasReflectableSeam's safe-query contract. 753/753 tests green.

PR #708, opened 2026-07-08
tilemap: render as world-space background (pre-sprite + camera transform) — v1.76.0

apotema:
## Summary
Closes the two T2 tilemap-rendering gaps the **pack-colony-demo** exposed (engine 1.75.1): the tilemap pass ran POST-sprite and in SCREEN space, so opaque terrain painted OVER the gameplay sprites and needed a baked camera offset to line up. Both are now fixed — tilemaps render as a proper **world-space background**.

### 1. Pre-sprite ordering (background, not overlay)
`game/loop_mixin.zig` `render()` now runs `renderTilemaps()` **before** `renderer.render()`, so terrain draws UNDER the houses/workers/hazard sprites. Gizmos still render last (sprite/gizmo order unchanged relative to each other).

### 2. World-space camera transform (pans/zooms with the world)
`game/tilemap_mixin.zig` `renderTilemaps()` wraps the pass in the SAME `camera.begin()/end()` sprites render through — gated on the renderer exposing a `begin()/end()` camera (`Game.CameraType != void`). Each map still draws with `camera_x/camera_y = 0`: the camera **matrix** (not a software offset) applies the pan/zoom, so a `Tilemap` entity's `Position` is a true world position that stays aligned with sprites at the same world coords under any pan/zoom. A camera-less renderer (headless/stub) folds the wrap away and keeps the T2 screen-space fallback (identity for a static view). The pre-camera world→screen Y offset (F3 `core.toScreenY`) is unchanged.

## Tests
- Flipped the ordering assertion to **background-first** (tile draws precede the sprite sentinel, which is now last).
- New camera-capable mock (`MockCamera` → backend `beginMode2D`) + a test asserting the pass runs INSIDE the camera transform and that a **camera pan shifts the recorded transform** while per-tile dests stay in world space.
- Existing F1/F3/F4 tilemap tests stay green.
- `zig build test` → 754/754 pass; `zig fmt` clean; files <1000 lines.

## Version
`build.zig.zon` bumped to **1.76.0**.

## Demo proof
Verified on pack-colony-demo (world Tilemap Position corrected to its true origin `{0,0}`): terrain renders as background with the 3 houses + 5 workers + hazard visible on top at correct positions.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,7 +1,7 @@
 .{
     .fingerprint = 0xe8a81a8dc7f48c87,
     .name = .engine,
-    .version = "1.75.1",
+    .version = "1.76.0",
     .minimum_zig_version = "0.16.0",
     .dependencies = .{
         // Requires labelle-core >= v1.20.0 — font_types aliases core's
--- a/src/game/loop_mixin.zig
+++ b/src/game/loop_mixin.zig
@@ -223,14 +223,17 @@ pub fn Mixin(comptime Game: type) type {
             // is suppressed, and only for the few frames the re-decode
             // takes. The common path (no gate armed) is unchanged.
             if (self.post_load_render_gate == null) {
-                self.renderer.render();
-                // Tilemap POST-SPRITE pass (T2 Phase 2): the engine owns
-                // pass ordering — entities first, tilemaps after. No-op
-                // when no Tilemap entities exist / renderer lacks the seam.
-                // Gated by the same post-load render gate as the world
-                // draw so restored tilemaps don't flash before their
-                // tileset textures re-bind.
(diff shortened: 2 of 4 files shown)


apotema:
Both addressed in 27bd12a:
- **@gemini** — moved the `tilemaps.count()==0` fast-path BEFORE `reapGhostTilemaps` so tilemap-less games skip the reap iterator entirely each frame (reap only removes, so an empty table has nothing to reap); kept a post-reap check for the all-ghosts case.
- **@codex** — the single-camera limitation is real for split-screen; documented it at the camera wrap and filed **#709** (draw the background per active camera). Single-camera — the common case and the colony demo — is correct. Deferred rather than duplicating the renderer's per-camera loop in this PR.

PR #712, opened 2026-07-08
tilemap: per-camera unbound background via renderWithLayerHooks — closes #709 — v1.79.0

### LINKED ISSUE #709: Tilemap background: draw through every active camera (split-screen / multi-camera)
T2 tilemaps render as a world-space background via a single `getCamera().begin()/end()` wrap in `renderTilemaps` (engine v1.76.0, #708). Under split-screen / multi-camera, `GfxRenderer.render()` draws sprites once PER active camera (each viewport), but the tilemap background pass runs once with the primary camera — so secondary viewports show the primary camera's terrain (or misaligned terrain) behind their sprites.

Fix: run the tilemap pass per active camera, matching the renderer's camera loop + viewports. Needs the engine to iterate the camera manager's active cameras for the background pass (or a renderer hook) without breaking the engine↔gfx separation (engine orchestrates ordering; gfx doesn't know about tilemaps).

Single-camera (the common case + pack-colony-demo) is correct today. Flagged by codex on #708.

apotema:
## What

Follow-up to T3 (#711) that makes the UNBOUND tilemap **background** per-camera, closing **#709**. T3 already drew *bound* `.tmx` layers per active camera; the *unbound* background (name-unmatched / back-compat layers) was still drawn **once through the primary camera** before the per-camera hook loop, so under split-screen only the primary viewport showed it.

Uses gfx **v1.24.0**'s dual-hook `renderWithLayerHooks(on_before_layers, on_after_layer)` — `on_before_layers` fires once per active camera, inside that camera's transform + **viewport scissor**, before the sprite stack (the missing per-camera pre-sprite slot the engine couldn't reproduce, since `applyViewport` is gfx-private). Bumps engine to **v1.79.0**.

## The before-hook background wiring

- New gate `Game.tilemap_percamera_background_supported` = the T3 interleave gate **AND** `@hasDecl(RenderImpl, "renderWithLayerHooks")`.
- `loop_mixin.render` three-tier branch:
  1. **per-camera** (gfx ≥1.24.0): one `renderWithLayerHooks(*Game, self, tilemapBackgroundHook, tilemapLayerHook)` call — both tilemap draws ride gfx's own per-active-camera loop (scissor included). `tilemapBackgroundHook` (on_before_layers) draws the **unbound** layers per camera, culled to that camera's world rect via the existing `cameraCullRect(cam)`; `tilemapLayerHook` (on_after_layer) draws **bound** layers, unchanged.
  2. **middle tier** (older single-hook `renderWithLayerHook`, gfx 1.22–1.23): single-primary `renderTilemapBackground` + `renderWithLayerHook` (bound layers still per-camera).
  3. **T2 fallback** (no hook): whole-stack pre-sprite background + `render()`.
- Shared `drawUnboundLayers(self, cam)` backs both the per-camera hook and the legacy single-primary pass, so they can't diverge.
- `cameraCullRect` hardened: `getViewport` is referenced only inside a comptime-gated block, so a hooks-capable renderer without a viewport still compiles (screen-size origin-cull fallback). Removed the single-primary-camera "limitation" doc — the background is now genuinely per-camera.

## #709 fully addressed

Bound layers were already per-camera (T3 #711). This closes the remaining gap: **unbound background layers now render once per active camera, each scissored + culled to its own viewport rect.** Verified by the new split-screen background test.

## Tests (interleave suite → 11)

- **`#709 split-screen: UNBOUND background renders PER active camera, culled to its rect`** — a 400×50 (taller-than-screen) map with a single `ground` (unbound) layer and two cameras at x≈100 and x≈5000. Asserts BOTH viewports draw the background and each is culled to its OWN region (camera 1 near x=5000, not the origin). **Proven to catch the bug**: forcing the middle-tier (single-primary) path makes exactly this test fail; the per-camera path passes.
- **`gate: a renderWithLayerHook-only renderer keeps interleave but NOT per-camera background`** — an `OldHookRender` mock (old single hook, no `renderWithLayerHooks`) locks the tiering (`interleave = true`, `percamera = false`).
- The mock now exposes `renderWithLayerHooks`, so the existing interleave / bound-cull (#711 P1) / #709-count / save-load / comptime-gate tests all re-run on the per-camera path and stay green.

Full `zig build test` green (Zig 0.16.0); `zig fmt`-clean; files <1000 lines. Engine at 1.79.0.

Closes #709.

Do not merge.


apotema:
Fixed in a680428 — thanks @gemini. The single-primary `renderTilemapBackground` now gates the `getCamera()` arg with `camera_capable` (passing `undefined` otherwise; `cameraCullRect` is gated on `camera_cullable` and returns an empty rect without dereferencing), matching the already-gated begin/end. 773/773 green.

Issue #723, opened 2026-07-08
Camera-bound layers: bind a render layer to a specific camera


apotema:
## Summary

Let a render **layer** be bound to a **specific camera** (or to none), instead of today's fixed binary: world-space layers are transformed by the one global camera, screen-space layers are pinned. This is a natural follow-on to the camera-prefabs MVP (#714), which made cameras first-class entities.

## Motivation

`LayerConfig.space` (labelle-gfx `src/layer.zig`) currently conflates two orthogonal questions:

1. **Whose transform applies?** — `.world` → the single active camera; `.screen` / `.screen_fill` → none (pinned).
2. **Fit or fill?** — `.screen_fill` stretches to the framebuffer; everything else aspect-fits.

Splitting out #1 — an explicit per-layer camera binding — unlocks three things people already want, with one concept:

- **Parallax.** Bind a backdrop layer to a camera that follows the main camera at a fraction (e.g. 0.4×). Today we hand-pin skies as `screen`/`screen_fill` and get no depth. (This came up directly while fixing the Flying-Platform sky.)
- **Multi-camera** — minimap / split-screen / picture-in-picture. A layer bound to an overview camera, combined with the `Camera` component's existing `viewport: ?Viewport` sub-rect, is split-screen for free.
- **GUI ↔ world decoupling.** UI layer bound to *no* camera (pinned HUD) while the world layer follows the gameplay camera — the "GUI shouldn't move with the camera" behavior, made explicit rather than implicit in `space`.

## Proposed approach

- Add an optional camera **tag** to `LayerConfig` (layers are comptime; cameras are runtime entities, so the binding is by tag, not entity ref).
- `Camera` entities carry a matching tag. The renderer resolves tag → active camera once per layer per frame; layers with no tag stay pinned.
- `.world` maps to a default `"main"` tag → **zero migration** for existing games.
- Culling becomes **per-camera** (each bound camera has its own cull viewport).

## Scope (cross-repo)

- **labelle-gfx** — binding field on `LayerConfig`; `retained_engine` sets the per-layer camera before its draw calls; per-camera cull viewport.
- **labelle-engine** — camera registry keyed by tag (tag → `Camera` entity); feed the resolved active camera per layer into the render pipeline.

Full design, API sketch, alternatives, and phasing in **`RFC-CAMERA-LAYER-BINDING.md`**.

## References

- Camera-prefabs MVP: #714
[...]

Issue #714, opened 2026-07-08
Camera component + studio editor bridge (camera-prefabs MVP)


apotema:
Implements the engine half of the **Camera Prefabs** RFC (`RFC-CAMERA-PREFABS.md`, branch `rfc/camera-prefabs`, PR labelle-toolkit/labelle-engine#715): a built-in `Camera` component that makes the game camera an authored, studio-manipulable entity — **without** migrating any game's camera script (soft ownership; the component is a seed, not a runtime dictator).

> Updated to fold in the maintainer's five reviewed refinements. **A/B/C are engine-side and are reflected in the checklist below.** **D** (gizmo screen-size dependency / gfx#249) is studio-side — labelle-toolkit/labelle-studio#67. **E** (default-insertion is a runtime `ensureDefaultCamera` helper the assembler calls) is assembler-side — labelle-toolkit/labelle-assembler#569.

### Why

Today the camera is gfx-owned, not an ECS entity: `game.CameraType` just re-exposes `RenderImpl.CameraType` (`src/game.zig:1264–1266`), and there is no `Camera` component. So the studio digest (`editor_scene_digest`, `src/editor_api.zig:296`), which only publishes `Position` entities, cannot see the camera at all, and the only per-component edit path is `editor_set_entity_position` (`src/editor_api.zig:274`). See the RFC "Current state" section for full citations.

### Scope — single-camera MVP

- [ ] **`Camera` built-in component.** New `src/camera.zig`-style module + `Game.CameraComp`, modeled on `TilemapComp` (`src/game.zig:225–229`, `src/tilemap.zig` — "Engine built-in, NOT a `ComponentRegistry` component"). Shape: `zoom: f32 = 1.0`, `viewport: ?CameraViewport = null`. **(refinement)** `viewport` uses an **engine-local** rect type — the engine is renderer-agnostic and gfx is an optional/test-only dep, so the component must **NOT** reference gfx's `ScreenViewport`; define a tiny engine-local `{ x, y, width, height }` mirror (inert in MVP anyway). Position (xy) comes from the entity's `Position`. Reserve room for a future `follow`.
- [ ] **Scene-loader apply.** A `"Camera"` branch in `src/jsonc/component_apply.zig`, guarded `comptime !Components.has("Camera")` exactly like the Tilemap branch (`:126–133`), so a project-registered `Camera` still wins.
[...]

PR #724, opened 2026-07-08
docs: RFC for camera-bound layers (#723)

apotema:
Design doc (**Draft**) for **camera-bound layers** — the design phase of #723.

Binds a render layer to a specific camera by **tag**, generalizing today's world-vs-screen binary into a transform axis that's orthogonal to fit. One concept unlocks:

- **parallax** — a backdrop layer bound to a camera that tracks the main one at a fraction;
- **multi-camera** — minimap / split-screen / PiP (with the `Camera` component's existing `viewport` rect);
- **pinned HUD** — a UI layer explicitly decoupled from a world-following camera.

Layers stay comptime and cameras stay runtime entities by binding via a **tag** rather than an entity id; `.world` layers default to a reserved `"main"` tag so existing games need zero edits.

Follow-on to the camera-prefabs MVP (#714). No code in this PR — it's the RFC for review. Implementation is phased: gfx `LayerConfig` field + per-layer `setCamera` → engine tag→camera registry → split-screen/gizmo.

Issue: #723

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

### DIFF
--- a/RFC-CAMERA-LAYER-BINDING.md
+++ b/RFC-CAMERA-LAYER-BINDING.md
@@ -3,15 +3,15 @@
 **Issue:** labelle-toolkit/labelle-engine#723  
 **Status:** Draft  
 **Author:** Alexandre  
-**Date:** 2026-07-08
+**Date:** 2026-07-08 (rev 2, 2026-07-09 — verified current architecture across gfx/engine/backends)
 
 ## Problem
 
-The camera-prefabs MVP (#714) made cameras first-class entities — you can author a `Camera` component, seed it on load, and manipulate it. But at *render* time there is still exactly one camera that matters: the renderer applies a single active camera to every world-space layer and no camera to screen-space layers.
+The camera-prefabs MVP (#714) made cameras first-class entities — you can author a `Camera` component, seed it on load, and manipulate it. But a layer cannot choose *which* camera transforms it: world layers follow **the** camera, screen layers follow none.
 
-`LayerConfig.space` (labelle-gfx `src/layer.zig`) conflates two orthogonal axes:
+`LayerConfig.space` (labelle-gfx `src/layer.zig:20`) conflates two orthogonal axes:
 
-1. **Whose transform applies?** — `.world` → the one active camera; `.screen` / `.screen_fill` → none (pinned).
+1. **Whose transform applies?** — `.world` → the active camera; `.screen` / `.screen_fill` → none (pinned).
 2. **Fit or fill?** — `.screen_fill` stretches to the framebuffer; `.screen` / `.world` aspect-fit.
 
 Because "which camera" is welded to "world vs screen," a game cannot:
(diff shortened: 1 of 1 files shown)


apotema:
Rev 4 (ea709b9) folds in the review findings. Disposition of every thread, newest-first:

**Fixed in rev 4:**
- **Sync scripted secondary cameras during gameplay** (codex) — the biggest catch: the parallax script had no API to drive a secondary camera. Added `getCameraByTag(tag)`, extending #714 soft-ownership (component = seed; scripts drive the slot imperatively). New "Driving secondary cameras" section.
- **Clear stale secondary camera slots on reseed** (codex) — seeding is now specified reset-then-seed: deactivate slots 1–3 + clear the tag map before iterating.
- **Rebuild bindings after save-load** (codex) — `loadGameState` added to the reseed triggers (verified: `save_load/load.zig:584-610` reattaches without seeding).
- **`tag: []const u8` lifetime/ownership** (gemini, coderabbit) — new "Tag storage & serde" note: the tag vocabulary is comptime-closed (layer bindings + "main"), stored bounded/interned, never a heap slice; `applyCameraComponentJson`'s transient arena stays safe; digest/bridge/save round-trip called out as a contract bump.
- **Round-trip tags through built-in channels** (codex) — same note covers digest (`camera:{…}` gains `tag`), `editor_set_component`, save/load.
- **Fallback when no camera matches a tag** (codex) — the loop sketch now shows the `!rendered` fallback branch matching the prose (world → default cam, screen → pinned, warn once).
- **Preserve the per-camera pre-layer hook** (codex) — Phase 1 gfx bullet now requires `on_before_layers` (tilemapBackgroundHook) to keep firing once per active camera before its first world layer across the inversion.
- **Reserve only three slots for non-main** (codex) — overflow wording fixed: 3 non-`"main"` slots, warn on the 4th.
- **Represent tag bindings one-to-many** (coderabbit) — tag map specified as `tag → slot set` from day one; Phase 1 seeds one per tag, Phase 3 activates fan-out without changing the shape.
- **Minimap multi-pass model** (codex) — worked example now states the one-tag-per-layer boundary honestly (own-content icon layer works today; world-mirror minimap needs many-tags or overview-tagged-main, moved to Open questions).

**Already addressed by revs 2–3 (threads anchored to rev 1/2):**
- **Route null world bindings to main** (codex), **null resolve** (gemini) — rev 2's `implicitTag` + rev 3's default-camera invariant: `"main"` always resolves (slot 0 always active).
- **Resolve tags to a camera snapshot with position** (codex) — rev 4 also states it explicitly: resolution targets the gfx pool camera (seeded from `Position`, then driven via `getCameraByTag`); the engine component is never dereferenced at render time.
- **Keep main bound to the live renderer camera** (codex) — rev 3: an authored `"main"` camera *configures* slot 0; imperative drivers keep working; zero-camera scenes are first-class.
- **Duplicate-tag behavior** (coderabbit) — rev 3 + rev 4: uniform fan-out semantics, one-per-tag seeding in Phase 1.

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

apotema:
Rev 5 (8a70e07) — disposition of the round-2 findings on rev 4:

**Fixed:**
- **Conflicting Phase-1 duplicate-tag rules** (coderabbit) — real contradiction I introduced in rev 4: the seeding bullets seeded *every* non-main camera while the duplicate-tags rule said one-per-tag. Now uniform: **first camera per tag wins in Phase 1** (main and non-main alike), fan-out seeding is Phase 3.
- **Deterministic `getCameraByTag` for shared tags** (coderabbit) — specified: **lowest active slot wins** (stable; `"main"` ⇒ slot 0); plural accessor deferred to Phase 3.
- **Scene-loader parsing for tags** (codex) — the initial-authoring channel (`jsonc_scene_bridge`) added to the round-trip list alongside save/load, digest, and the bridge.
- **Digest views per tagged camera** (codex) — the digest's `view` must resolve against the camera's *own* slot, not slot 0; added to the serde note.
- **Preserve the after-layer hook for bound tilemaps** (codex) — Phase-1 gfx constraint now covers **both** hooks: `on_after_layer` fires per layer × camera pass, inside that camera's transform and fit mode.
- **Repeated `tag: []const u8` lifetime finding** (gemini) — the struct snippet now marks the field as a conceptual type with a pointer to the storage note (bounded/interned, never a heap slice), which has covered the arena-lifetime concern since rev 4.
- **Unbound screen layers under viewports** (codex) — the sketch's fallback comment now cross-references the pinned-HUD-under-split-screen Open question; the sketch shows the once-full-window baseline deliberately.

**No change:**
- **"Add a fallback when no camera matches a tag"** (codex) — the rev-4 sketch already tracks `rendered` and runs the unbound path when no camera matched, which is exactly the requested behavior; looks like a re-post of the rev-3 finding.

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

PR #727, opened 2026-07-10
feat(camera): tagged multi-camera seeding + getCameraByTag (camera-bound layers RFC #723, PR2)

apotema:
**Phase-1 PR 2** of the Camera-Bound Layers RFC (#723 / RFC #724). Adds the engine half: `Camera.tag`, tagged multi-camera reset-then-seed, a `getCameraByTag` accessor, and tag round-trip through every Camera channel. Pins **gfx v1.26.0** (PR1, merged + released).

## Changes
- **`Camera.tag`** (`src/camera.zig`): inline sentinel-terminated `[16:0]u8`, default `"main"`, never a heap slice (the `applyCameraComponentJson` transient-arena constraint). Helpers `makeTag()`/`tagSlice()`/`setTagSlice()` (truncate at 15 to match gfx's `tag_buf`).
- **`seedCameraFromComponent` → reset-then-seed** (`camera_mixin.zig`): `getCameraManager().resetSecondary()` first, then iterate ALL `{Position, Camera}`. `"main"` → slot 0 (first wins); other tags → first-per-tag claims next free slot 1–3. Warn-once (latching) on unknown tag vs the comptime `LayerEnum` `.camera` vocabulary, extra main, duplicate tag, slot overflow. Zero cameras → slot 0 untouched (invariant).
- **`getCameraByTag(tag) ?*CameraType`** (`misc_mixin.zig`/`game.zig`): forwards to `findByTag` (lowest active slot); comptime null no-op on camera-less renderers.
- **`loadGameState` reseed**: re-seed after Camera components reattach (tagged cameras in a save would otherwise come back with dead slots).
- **Channels** (tag round-trips through all): scene/prefab JSONC (`component_apply.zig`), save/load serde, studio digest (adds `tag`, resolves `view` against the camera's own slot), `editor_set_component`. Editor-bridge contract doc bumped **v1.5 → v1.6**.

## Tests
+11 in `test/editor_api_test.zig`: multi-camera slot assignment, reseed-clears-removed, **zero-camera no-op invariant**, first-per-tag warn, overflow, `getCameraByTag` lowest-slot, load-path reseed, JSONC + set_component + digest tag round-trip, bounded-buffer truncation. The `CamRender` mock's `CameraManagerType` was upgraded from `struct{}` to a 4-slot tag manager mirroring gfx's contract. `zig build test` exit 0 against gfx v1.26.0.

Note: the engine is renderer-agnostic (gfx camera types arrive via `RenderImpl.CameraManagerType`), so tests exercise the manager contract through the in-tree mock; the real gfx↔engine integration is proven by FP PR4. Releases as engine **1.83.0** (follow-up release commit).

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -58,9 +58,17 @@
         // hand-rolled stand-in. `lazy` so downstream consumers of the engine
         // module never fetch it — it is reached only by the `test` step via
         // `b.lazyDependency`, mirroring `zspec`.
-        // DEV PIN — gfx#724 PR1 (camera-bound layers). Flip to gfx v1.26.0 tag
-        // at release. `lazy` is dropped: a `.path` dep is local, not fetched.
-        .labelle_gfx = .{ .path = "../labelle-gfx-724" },
+        // Bumped to gfx v1.26.0 for camera-bound layers (RFC #723): engine
+        // 1.83.0's tagged-camera seeding drives gfx's CameraManager
+        // setTag/findByTag/resetSecondary (added in 1.26.0). Test-only + lazy —
+        // the engine module takes no direct gfx dep (the renderer arrives via
+        // RenderImpl), so the effective gfx floor is enforced at the consuming
+        // game's build; this pin only feeds the tilemap sub-package test.
+        .labelle_gfx = .{
+            .url = "https://github.com/labelle-toolkit/labelle-gfx/archive/refs/tags/v1.26.0.tar.gz",
+            .hash = "labelle_gfx-1.26.0-nMpCPMYYBQBBX_7Js1AHBQKWYeI7Hm8U2aEL8t9cOIJO",
+            .lazy = true,
+        },
     },
     .paths = .{
         "build.zig",
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -58,11 +58,9 @@
         // hand-rolled stand-in. `lazy` so downstream consumers of the engine
         // module never fetch it — it is reached only by the `test` step via
         // `b.lazyDependency`, mirroring `zspec`.
-        .labelle_gfx = .{
-            .url = "https://github.com/labelle-toolkit/labelle-gfx/archive/refs/tags/v1.24.0.tar.gz",
-            .hash = "labelle_gfx-1.24.0-nMpCPF3VBAANEcKU73ait49fJJ2rchmMJgsAW8gI2ntB",
-            .lazy = true,
-        },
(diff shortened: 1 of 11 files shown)


apotema:
Thanks @gemini-code-assist.

- **save.zig HIGH (unescaped tag)** — fixed in `d454c3c`: the tag now serializes via the existing `writeJsonString` escape helper (like every other string field), so a tag with `"`/`\` can't corrupt the save. Added a round-trip test with a tag carrying both a quote and a backslash (`a"b\c`) that would break on raw output and now survives save→load intact.
- **camera_mixin.zig `inline for` MEDIUM** — declining. The vocabulary is the tiny set of authored layer `.camera` tags (a handful), and `inline for` over a comptime enum inside a runtime fn is the idiomatic pattern here; the unroll/bloat cost is negligible at this size.

`zig build test` exit 0 on `d454c3c`.

apotema:
Thanks @chatgpt-codex-connector — dispositioning all four:

- **(#1) Gate tagged camera-manager calls** — fixed in `058ddd3`. Added `hasTaggedCameraManager(G)` (verifies the manager declares `resetSecondary`/`setTag`/`setActive`/`getCamera`/`findByTag`). `seedCameraFromComponent` splits at comptime: tagged manager → the multi-slot seeding; non-tagged manager → the pre-PR single-camera fallback (no reset/tags), so an old `struct{}`-style manager compiles and behaves exactly as before. `getCameraByTag` also gates on it (comptime null no-op). Test: an `OldManagerGame` mock with a non-tagged manager proves the seeder compiles + does the single-camera fallback.
- **(#3) Seed slot 0 directly** — fixed in `058ddd3`: the `"main"` path now writes via `getCameraManager().getCamera(0)` so the transform and the `"main"` tag always land on the same slot regardless of what `getCamera()` selects.
- **(#2) Escape camera tags when saving** — already fixed in `d454c3c` (you reviewed the pre-fix head 8292aa0): serializes via `writeJsonString` now, with a quote+backslash round-trip test.
- **(#4) Zero-camera reset / soft-ownership** — declining for Phase 1. reset-then-seed clearing stale secondaries is intentional (RFC): secondary cameras are *authored* (not pure-script) in Phase 1, so a zero-authored-camera scene legitimately has no secondaries to preserve, and the re-seed fires only on scene-load / editor apply-while-paused, not per frame. The soft-ownership refinement for script-driven secondaries under editor-pause is a Phase-2/3 consideration.

`zig build test` exit 0 (821/821). New HEAD `058ddd3`.

Issue #738, opened 2026-07-10
Language plugins P1: labelle-scripting repo bootstrap — shared glue + lua sub-module


apotema:
Phase 1 of the language-plugins epic. Epic: #237 · RFC: labelle-toolkit/labelle-engine#730 (`RFC-LANGUAGE-PLUGINS.md` rev 11) · POC: labelle-toolkit/labelle-engine#734 (`spike/language-plugins/`)
**Target repo: `labelle-scripting` (to be created — tracked here until bootstrapped).**

Bootstrap the single language-plugin repo: shared glue + the `lua` sub-module.

- **Shared glue**: contract binding; script discovery (language convention dirs; `@embedFile` for release, disk sources in dev); the plugin controller (setup/tick/deinit in the two-block order); the event dispatcher (subscription registry + poll drain → handler callbacks).
- **lua sub-module**: vendored Lua 5.4 via `b.lazyDependency` (POC recipe: ~20 hand-declared externs, no `@cImport`); `labelle.*` bindings + the #237 reference sugar — `game:query(...)` generic-for iterator, entity get/set as tables, `labelle.on` event handlers.
- Singular `.language` declaration consumed from project config (validation itself is the assembler ticket).

**Acceptance**: attached to a demo project, `.lua` scripts drive entities through init/update/deinit; events work both directions; an unchosen language's runtime is never fetched (lazyDependency verified).

Issue #237, opened 2026-01-13
feat: Lua scripting module (labelle-lua)


apotema:
## Summary

Add optional Lua scripting support via a new `labelle-lua` plugin. This enables:
- Rapid iteration without recompilation
- Mod support for games
- Accessible scripting for non-Zig developers
- Hot reload during development

## Motivation

While Zig scripts offer maximum performance, many game logic scenarios benefit from a dynamic scripting language:
- **AI behaviors** - Frequently tweaked, rarely performance-critical
- **Game rules** - Designer-facing, needs fast iteration
- **Dialog/cutscenes** - Data-driven, benefits from hot reload
- **Mod support** - Safe sandboxed environment for user content

Lua is the industry standard (World of Warcraft, Roblox, Love2D, Defold) with:
- Tiny runtime (~200KB)
- Fast FFI boundary crossing
- LuaJIT option for near-native performance
- Battle-tested in AAA games

### Plugin Structure

\`\`\`
labelle-lua/
├── src/
│   ├── lua_engine.zig      # Lua VM management
│   ├── bindings/
│   │   ├── game.zig        # game.* API
│   │   ├── entity.zig      # Entity manipulation
│   │   ├── input.zig       # Input bindings
│   │   ├── math.zig        # Vec2, transforms
│   │   └── components.zig  # Component access
│   ├── script_adapter.zig  # Adapts to ScriptFns interface
│   └── hot_reload.zig      # File watching + reload
├── build.zig
└── build.zig.zon
\`\`\`

### Core API

\`\`\`zig
pub const LuaEngine = struct {
    state: *LuaState,
    allocator: std.mem.Allocator,
    scripts: std.StringHashMap(ScriptHandle),
    
    pub fn init(allocator: std.mem.Allocator) !LuaEngine;
    pub fn deinit(self: *LuaEngine) void;
    
    // Script loading
    pub fn loadScript(self: *LuaEngine, name: []const u8, path: []const u8) !void;
    pub fn loadScriptString(self: *LuaEngine, name: []const u8, source: []const u8) !void;
    
    // ScriptFns-compatible callbacks
    pub fn callInit(self: *LuaEngine, script: []const u8, game: *anyopaque, scene: *anyopaque) void;
    pub fn callUpdate(self: *LuaEngine, script: []const u8, game: *anyopaque, scene: *anyopaque, dt: f32) void;
[...]

PR #749, opened 2026-07-10
feat(#737): Script Runtime Contract v1 — C-ABI surface for language plugins

### LINKED ISSUE #737: Language plugins P1: Script Runtime Contract v1 (C-ABI, comptime-gated)
Phase 1 core of the language-plugins epic. Epic: #237 · RFC: labelle-toolkit/labelle-engine#730 (`RFC-LANGUAGE-PLUGINS.md` rev 11) · POC: labelle-toolkit/labelle-engine#734 (`spike/language-plugins/`)

Implement the **Script Runtime Contract v1** in the engine: the versioned C-ABI surface every language binds.

- entity create/destroy/find, prefab_spawn; **component get/set/has/remove BY NAME as JSON** — generalizing `editor_set_component`'s serde-reflection dispatch (`src/editor_api.zig:327`) minus the editor gating, plus the read twin;
- **events**: subscribe + poll-drain bridging `GameEvents` (FIFO per subscriber — semantics pinned by the POC's `contract/contract.h`);
- **query**: component names → entity-id snapshot over ECS views; input, time_dt, scene_change, log;
- **comptime-gated**: compiles only when a language plugin is attached — zero size/codegen impact otherwise (gate test required); versioned like the editor contract; main-thread-only v1; JSON encoding v1 (tagged-value later).

**Acceptance**: Zig contract test suite exercising every function against a real game instance; a no-language-plugin game builds byte-identical; the POC behaviors replayable against the real engine.

apotema:
**Script Runtime Contract v1** — the C-ABI surface every scripting language binds. Closes #737 (Phase 1 core of the language-plugins epic #237; design: `RFC-LANGUAGE-PLUGINS.md` #730; shapes proven by the POC #734).

## Architecture (editor_api-mirrored)

- `bind(&g)` once from the generated main → comptime `Holder` publishes a **type-erased vtable**; the 15 `export fn labelle_*` symbols carry no comptime types. Pre-bind, every export is a safe no-op.
- **Zero cost when unused**: reachable only through the lazily-analyzed `root.script_contract` — a build that never references it emits no symbols (the same `nm`-verified gate `editor_api` uses).
- Generated-main touchpoints documented for the assembler: `bind(&g)` after init; `drainEvents(&g)` **after tick, before `dispatchEvents`** (dispatch swaps the event buffer out — tapping after it always sees an empty list; the ordering is load-bearing and documented).

## Surface

`contract_version · entity_create/destroy · prefab_spawn · component_set/get/has/remove · query · event_emit/subscribe/poll · scene_change · log · time_dt` — consumer header at `contract/labelle_script.h` (`LABELLE_CONTRACT_VERSION 1`).

Notable implementation decisions (all documented in-file):
- **Components by name** over the registered set via the scene-bridge dispatch (not editor_api's Camera allowlist — scripts are the game's own code); `Position` included and routed through `setPosition` so render dirty-tracking fires; slice fields use the scene bridge's `nested_entity_arena` lifetime convention; visual built-ins with renderer side-tables (Sprite/Shape/Camera/…) are explicitly out of v1.
- **Double-buffered emit arenas**: JSON-parsed event payloads with slice fields must survive until end-of-frame dispatch; a single reset-at-drain arena would UAF them. Two arenas flip per drain; pinned by a slice-bearing-payload test.
- **Poll semantics POC-pinned**: `"<name> <json>"` FIFO, truncated entry still consumed, 256-cap drop-newest, subscribe dedupes, rebind = fresh session.
- `time_dt` reads the frame-profiler ring × `time_scale` (no extra generated-main splice); `query` writes always-valid JSON with truncation rollback.

## Verification

**904/904 tests, 170/170 steps** (baseline 885/885) — 19 new tests: pre-bind no-ops, component round-trips incl. unknown-name, 2-component query + filter, emit-by-name into the real buffer, subscribe/poll FIFO + filtering, slice-payload lifetime, `GameEvents = void` and empty-registry (`engine.Game`) compile-and-degrade paths.

Epic: #237 · next consumer: labelle-scripting bootstrap (#738)

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -225,6 +225,13 @@ pub fn build(b: *std.Build) void {
         // these cover the FPS counter and the overlay-facing surface.
         "test/frame_profiler_test.zig",
         "test/inspector_overlay_test.zig",
+        // #737 — Script Runtime Contract v1: the `labelle_*` C-ABI
+        // surface language plugins bind (RFC-LANGUAGE-PLUGINS). Covers
+        // pre-bind no-op safety, the bind/vtable dispatch for every op,
+        // component serde round-trips over the registry, name-dispatched
+        // queries, event emit-by-name, and the subscribe/drain/poll FIFO
+        // (incl. the two-arena payload lifetime).
+        "test/script_contract_test.zig",
     };
 
     for (test_files) |test_file| {
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -79,5 +79,9 @@
         "codegen",
         "audio_backend",
         "spec",
+        // The Script Runtime Contract header (contract/labelle_script.h)
+        // ships with the package — native-compiled language plugins bind
+        // against it (#737).
+        "contract",
     },
 }
--- a/contract/labelle_script.h
+++ b/contract/labelle_script.h
@@ -21,8 +21,11 @@
  *     nothing written, 0 returned).
  *   - Structured payloads are UTF-8 JSON (encoding v1).
  *   - Components are addressed BY NAME over the game's own component
- *     registry (the same set JSONC scenes author), plus the built-in
- *     "Position" ({"x":…,"y":…}).
+ *     registry plus the engine built-ins JSONC scenes author: "Position"
+ *     ({"x":…,"y":…}) and the five scene built-ins "Sprite", "Shape",
+ *     "Tilemap", "Camera", "Image" — routed through the scene loader's
+ *     own apply machinery (see the support table before
+ *     labelle_component_set).
  *   - Events are addressed by the game's `GameEvents` union tag name
  *     (e.g. "turret__fired", "engine__tick"); payloads are the variant
  *     struct as JSON.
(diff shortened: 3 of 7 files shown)


apotema:
All 5 findings fixed in the latest push — each with a pinning test, negative-verified (reverting each fix fails exactly its test):

1. **Inbox slot reuse** → `compactInbox()`: pending tail slides to front at head ≥ 16 or ≥ half the list; capacity bounded under a standing backlog (100-frame test).
2. **Header packaging** → `"contract"` added to `.paths`.
3. **Idempotent remove** → `hasComponent` guard on both branches; absent-remove fires no `onRemove` (hook-counter test via the OnReadyBump pattern).
4. **NULL-able ABI** → convention made consistent header↔code: nullable JSON payload inputs (`?[*]const u8` + shared `optionalJson`), out buffers NULL-only-with-cap-0, and a bonus fix discovered here: a cap-0 poll previously **silently consumed an entry** — now a documented no-consume probe.
5. **dt consistency** → `labelle_time_dt_stamp(f32)` (plugin-only, once per tick); `labelle_time_dt` returns the stamp verbatim, so mid-tick `time_scale` changes can't skew script dt vs Zig scripts; profiler fallback when unstamped.

909/909 tests (+5).

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

apotema:
Round-2 findings all fixed in the latest push (each negative-verified):

1. **Read liveness** — get/has now guard `entityExists` like set/remove; pinned via a `LeakyReadEcs` fixture that serves dead rows the way a real sparse set can (the standard mock would mask the bug).
2. **No same-tick subscription replay** — subscribe lands in a pending set absorbed at drain end; effective for events emitted **after** the current tick's drain (documented module + header).
3. **Built-in parity** — `component_apply.zig`'s Sprite/Shape/Tilemap/Camera/Image branches extracted as `pub` fns and dispatched by the contract with the scene loader's own precedence gates. **Full support matrix in the header — set/get/has/remove/query for all five**, removes via the typed channels (renderer untrack / tilemap runtime free), gets via a comptime round-trip filter (handle fields omitted and re-derived on set). Registered same-name components keep precedence.

916/916 tests (+7 from baseline 909).

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

apotema:
One more contract refinement in the latest push, prompted by review on the consumer PR (labelle-toolkit/labelle-scripting#1): **`labelle_query` now returns bytes REQUIRED** (snprintf semantics) — silent truncation was undetectable. Writing behavior unchanged (valid-JSON prefix, last-whole-id rollback); `required > cap` is the truncation signal; NULL/cap-0 is a pure sizing probe. `component_get`/`event_poll` deliberately stay written-bytes — query is the lone unbounded-cardinality op (asymmetry documented in module + header). 916/916.

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

apotema:
Rounds 3+3b in the latest push (all negative-verified):

- **`component_get` → required-size, all-or-nothing write** (finishing what query started); **the poll probe now truly returns the next entry's size without consuming** (round-1's returned 0 — consumer review caught the gap). Final sizing model in one header section: required-size for the unbounded ops, written-bytes+probe for poll.
- **`.alloc_always` on component AND event-payload parses** — `alloc_if_needed` let unescaped slice fields borrow the caller's temporary buffer (dangling after return); poison-after-call tests pin the copies.
- **Trailing-garbage rejection** on built-in component sets (before any mutation; JSONC comment tails stay legal) + the single-document rule documented.
- **Void-payload accept set**: exactly empty/`{}`/`null`; anything else -1, nothing buffered.

**920/920 tests.** With the consumer PR (labelle-toolkit/labelle-scripting#1) green at 19/19 against the mirrored mock, the surface is stable across four review rounds.

https://claude.ai/code/session_01P7YLw4hXFCCaY2LAUt4G1j

PR #798, opened 2026-07-24
feat: make destroyEntity cascade panic-safe against already-dead descendants

apotema:
## What

`game.destroyEntity` cascades over a by-value snapshot of child ids, but a descendant can already be dead by the time the walk reaches it:
- a consumer destroyed some entities first, then cascades a surviving ancestor (the flying-platform room teardown case);
- the same id is reachable twice in the walk (a re-parent remnant / a child under two parents);
- a sibling's `entity_destroyed` hook killed it mid-cascade.

Every step past the recursion — backend destroy, the heap-backed `Children` `ArrayList` free, `detachFromParent`, tombstone, hooks, preview telemetry — would then double-free / re-emit / trap on a dead id. That's precisely why downstream games use a hand-rolled non-cascading teardown (`destroyEntityOnly` + explicit id-array walking) instead of just cascading.

## Fix

An idempotence guard — `if (!self.ecs_backend.entityExists(entity)) return;` — as the **first statement** of both public destroy entry points (`destroyEntity`, and for parity `destroyEntityOnly`). `entityExists` is the alive-set oracle `detachFromParent` already trusts; a dead entity has already freed its own `Children` allocation, so bailing is a leak-free no-op. No change to `destroyEntityOnly`'s contract (listed children still stay alive).

## Why it matters

Lets a consumer (flying-platform's `31_drain_pass` room teardown) drop the non-cascading `destroyEntityOnly` + id-array walk and just `destroyEntity` the room — which in turn makes the `reapOrphans` fish/critter backstops genuinely removable (parenting now means the cascade cleans them up).

## Tests

`test/destroy_unlink_test.zig` (+6): subtree with a pre-destroyed child, grandchild destroyed before its ancestor's cascade, direct double-destroy, an already-dead subtree with grandchildren (no double-free), a normal all-live destroy, and `destroyEntityOnly` on a dead entity. All under `testing.allocator` (leak = fail). `zig build test` green.

https://claude.ai/code/session_011szWvquoss1yNX7KWSKCaM

---
<a href="https://app.blacksmith.sh/labelle-toolkit/codesmith/labelle-engine/pr/798"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787494811&installation_model_id=427192&pr_number=798&repository=labelle-toolkit%2Flabelle-engine&return_to=https%3A%2F%2Fgithub.com%2Flabelle-toolkit%2Flabelle-engine%2Fpull%2F798&signature=2ff097e39ea78d2db67ba7077e5500a5d874750f0c655e2ba83529c87a38a2a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/game/entity_mixin.zig
+++ b/src/game/entity_mixin.zig
@@ -264,6 +264,17 @@ pub fn Mixin(comptime Game: type) type {
                     destroyEntity(self, child);
                 }
             }
+            // Re-check liveness after the cascade. The entry guard only proved
+            // `entity` was alive at the START; a child's synchronous
+            // `entity_destroyed` hook can call back into the Game and destroy
+            // THIS entity mid-cascade (e.g. a teardown hook that removes an
+            // owner/room when one of its children dies). If it did, the entity
+            // has already run its full destroy — including freeing its own
+            // `Children` allocation — so the tail below (preview, the
+            // `Children` free, backend destroy, tombstone, hooks) must NOT run
+            // again on a dead id or it double-frees / re-emits. (`kids` is our
+            // independent dupe, freed by the defer above regardless.)
+            if (!self.ecs_backend.entityExists(entity)) return;
             // Preview telemetry emits BEFORE the actual destroy so any
             // editor-side consumer can still introspect the entity from
             // a `getComponent` style API while reacting to the frame — so the
--- a/src/game/entity_mixin.zig
+++ b/src/game/entity_mixin.zig
@@ -228,6 +228,25 @@ pub fn Mixin(comptime Game: type) type {
         }
 
         pub fn destroyEntity(self: *Game, entity: Entity) void {
+            // Idempotence guard — make the cascade panic-safe against an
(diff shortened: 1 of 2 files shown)


apotema:
@codex review

apotema:
@codex review

PR #806, opened 2026-08-04
fix(scenes): warn on unknown pack-namespaced components, not just PascalCase (#803)

### LINKED ISSUE #803: Unknown-component warning is PascalCase-only — typo'd pack-namespaced keys silently no-op
Split from codex round-4 finding on #802 (P2, deferred per the 3-round review policy).

`applyComponent`'s unknown-name diagnostic (`uf.warnUnknownComponent`, RFC #596 axis 4) fires only for PascalCase names. A pack-namespaced key with a typo — `industry__Storag` for `industry__Storage` — starts lowercase, misses the registry, and is **silently ignored** in every override path: ordinary `overrides` maps, flat-form keys, and (since #802) `@`-target patches, where the target records a hit and the scene loads successfully with an ineffective patch.

This predates #802 — `@` merely inherits it — which is why it's an issue rather than a round-4 fix: the correct change (extend the unknown-name warning to `<prefix>__<Pascal>`-shaped keys) affects all loader paths, not just target overrides, and deserves its own focused PR + tests.

Sketch: in the `applyComponent` fallthrough, warn when the name is PascalCase **or** matches the namespaced shape (`target_overrides.isComponentKeyShape` is the exact predicate, added in #802 — promote it to `unified_format` and share).

Refs: #801, #802, labelle-toolkit/labelle-assembler#651 (adjacent silent-drop class).

apotema:
Closes #803 (split from codex round 4 on #802).

`warnUnknownComponent` (RFC #596 Axis 4) fired only for PascalCase names — but every pack component is `<prefix>__<Pascal>`-namespaced and starts lowercase, so a typo'd `industry__Storag` silently no-op'd in **every** override path: ordinary `overrides` maps, flat form, and `@`-target patches (where the target records a hit and the scene loads with an ineffective patch).

## Changes
- `isComponentKeyShape` promoted from `target_overrides.zig` to `unified_format.zig` and shared — one predicate for "could this key ever name a registered component" (nonempty prefix, PascalCase suffix after the last `__`).
- `applyComponent`'s unknown-name gate widened to it. Plain lowercase data-shaped keys (`just_data`) stay silent, exactly as before — they aren't the authoring-mistake shape the RFC targets.
- `unified_format` re-exported through `root.zig` so tests/tooling share the loader's shape rules (and the `alreadyWarnedKey` dedup probe, which is how the new tests observe warnings without log capture).

## Tests
`test/jsonc/unknown_component_warn_test.zig`: shape-predicate table; typo'd namespaced key warns; registered namespaced component applies **without** warning; data-shaped lowercase key stays silent. Full suite green.

No behavior change for healthy projects: the assembler registers every pack component, so the widened gate only fires where something was already silently broken.

---
<a href="https://app.blacksmith.sh/labelle-toolkit/codesmith/labelle-engine/pr/806"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788442872&installation_model_id=427192&pr_number=806&repository=labelle-toolkit%2Flabelle-engine&return_to=https%3A%2F%2Fgithub.com%2Flabelle-toolkit%2Flabelle-engine%2Fpull%2F806&signature=e61a5a4845953d0e96fd6cc1f8a21a2bd6cd8b6e904c3812a8cb8b3c0d0db5e2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -92,6 +92,7 @@ pub fn build(b: *std.Build) void {
         "test/jsonc/bridge_leak_test.zig",
         "test/jsonc/nested_lifecycle_test.zig",
         "test/jsonc/target_overrides_test.zig",
+        "test/jsonc/unknown_component_warn_test.zig",
         // C2 — a project-registered component named `Tilemap` must win over
         // the engine built-in in the scene loader (no silent shadowing).
         "test/jsonc/tilemap_precedence_test.zig",
--- a/src/jsonc/component_apply.zig
+++ b/src/jsonc/component_apply.zig
@@ -204,16 +204,19 @@ pub fn ComponentApply(comptime GameType: type, comptime Components: type) type {
                 }
             }
 
-            // RFC #596 Axis 4: unknown PascalCase keys on an entity
-            // are treated as components, but we warn-once so typos
-            // (`Posiiton`) surface visibly. Lowercase names that
-            // reach here (e.g. legacy embedded structural keys that
-            // bypassed the structural / component split) are
+            // RFC #596 Axis 4: unknown component-shaped keys on an
+            // entity are treated as no-ops, but we warn-once so typos
+            // surface visibly — PascalCase (`Posiiton`) AND
+            // pack-namespaced (`industry__Storag`, #803: every pack
+            // component starts lowercase, so PascalCase alone missed
+            // the entire namespaced registry). Other lowercase names
+            // that reach here (e.g. legacy embedded structural keys
+            // that bypassed the structural / component split) stay
             // silently ignored — they're not authoring mistakes the
             // RFC catches. Position / Sprite / Shape are handled
             // above and returned before reaching this gate, so the
             // built-in components don't false-warn.
-            if (uf.isPascalCase(name)) {
+            if (uf.isComponentKeyShape(name)) {
                 uf.warnUnknownComponent(game.log, name);
             }
         }
--- a/src/jsonc/scene_loader/entity_walker.zig
+++ b/src/jsonc/scene_loader/entity_walker.zig
@@ -141,6 +141,11 @@ pub fn EntityWalker(comptime GameType: type, comptime Components: type, comptime
             // Fold enclosing references' matching `@` patches onto
             // this entity's own patch, outermost author last (wins).
             const fold = try to.foldMatches(inherited_targets, own_ref, parts.components, merge_arena.allocator());
+            // Removal-landing aggregation (#806): record whether THIS
+            // entity carried each matched removal's component; the
+            // declaring reference diagnoses never-landed removals
+            // once, after its whole fan-out resolved.
+            to.noteRemovalsCarried(fold, prefab_components, parts.components);
             const scene_components = fold.patch;
 
             // RFC #562 `null`-removal is scoped to reference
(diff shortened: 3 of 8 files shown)


apotema:
Round 1 pushed (3fe2a89) — all 4 findings addressed:

- **Flat-form classification** (codex P1 + CodeRabbit): `isFlatComponentKey` now delegates to `isComponentKeyShape`. This is quietly the biggest change in the PR: a **registered** flat-form namespaced component (`industry__Storage` authored flat on an entity) previously never attached — `synthesizeFlatComponents` dropped it before the loader saw it — and now applies. Capability test added alongside the flat-typo-warns test.
- **No-op removal warning** (codex P2): new `warnNoopRemoval` fires when a `null` removal matches neither the resolved prefab nor the entity's own pre-fold patch (typo'd removals previously read as success); wired into both loader removal branches, with a legit-removal-stays-silent test.
- **Direct re-export** (CodeRabbit): `engine.isComponentKeyShape` per the root.zig convention.

Cross-repo note: the assembler's classifiers don't yet accept flat namespaced keys (top-level `UnknownSceneKey` false rejection, hybrid gate, lint, pass-1 migration) — filed as labelle-toolkit/labelle-assembler#652 for the same treatment `isTargetKey` got in #650.

Full suite green.

apotema:
@codex review

apotema:
Round 2 pushed — all 3 codex P2s addressed:

- **Inner-adds/outer-removes** — `FoldResult` now carries the matched patches (inner→outer); the no-op check consults them, so a higher-precedence removal of an inner-@-added component is recognized as matching something real. Test covers the wrapper-prefab construction.
- **Duplicate warnings** — removal entries the reference authored itself are owned entirely by the specialized nested-component walker (which emits the `@ref`-guidance variant for nested-carried keys, and now the generic no-op message for keys carried nowhere); the generic path only handles fold-contributed removals. Dedup probe keys follow the split.
- **Lowercase removal keys** — `warnNoopRemoval` gates on `isComponentKeyShape`, so a `null` value can't change the classification of data-shaped keys. Test added.

Full suite green.

apotema:
Round 3 pushed — the missed fourth finding from the 13:59 review batch (fan-out removal false no-op) is fixed by aggregation: `TargetEntry` tracks per-removal-key landing across every matched entity (`noteRemovalsCarried` after each fold, counting prefab, own pre-fold patch, and other matched patches' contributions), and `checkAllMatched` diagnoses never-landed removals once per declaring reference. This subsumes the per-entity contributed-removal warning entirely. Tests: partial-fan-out removal stays silent; nowhere-landing removal warns once.

Per the review policy that's the final fix round — merging on green once verdicts settle; anything new gets an issue.

apotema:
CodeRabbit's final pass found the same dedup-probe mismatch codex did (tracked in #807, now extended with its adjacent collector-filtering Minor). Both are post-round-3 diagnostics-precision items in the warning path only — no load-behavior impact — so per the review policy they ride #807 rather than a 4th round. Merging on green.

Issue #810, opened 2026-08-07
Game constants remainder: finish FP library tuning-value migration


apotema:
**RFC:** `RFC-CONSTANTS.md` (in this repo)
**Sibling:** #809 (i18n) — shares the scan-and-codegen machinery

## Problem

Tuning values live wherever the code reading them happens to live. `flying-platform-labelle` has **38 such constants across 17 files**. They aren't hard to find — nothing relates them to each other.

```
packs/rooms/…/26_construction_progress.zig      BUILD_TIME              = 5.0
packs/rooms/…/32_carcase_build_dispatcher.zig   CONSTRUCTION_DURATION_S = 5.0
packs/rooms/…/40_repair_progress.zig            REPAIR_TIME             = 5.0
packs/rooms/…/40_repair_progress.zig            PER_PIECE_TIME          = 5.0
packs/rooms/…/29_deconstruction_progress.zig    DECONSTRUCT_TIME        = 4.0
```

Five timings, four files, four sharing `5.0` with no record of whether that's intent or coincidence. `SHIP_SPEED` is defined twice with different values (280 in `packs/transport`, 150 in `packs/combat`) — legitimate, but nothing says so and nothing would catch it if one were meant to track the other.

Decay rates cluster in `libs/needs_machine/src/config.zig`, which is the closest thing to this proposal already — and shows the failure mode. It holds `health_drain_rate = 0.0`, a value the game's CLAUDE.md spends four paragraphs warning about ("*when diagnosing a stalled colony, check for needs pinned at 0 instead of waiting for deaths*"). That warning exists because the value is invisible where it sits.

## Proposal

`constants/` at project root, one YAML file per domain, scanned like `scripts/` and `locales/`:

```yaml
# constants/construction.yaml
build_time: 5.0        # seconds
repair_time: 5.0
deconstruct_time: 4.0
```

Nesting maps onto the accessor path → `C.decay.hunger.rate`. Packs use `packs/<name>/constants/*.yaml` with `<pack>__` namespacing.

### Comptime via codegen, not a comptime YAML parser

"Comptime only" must not be read as `@embedFile` + comptime parse — YAML is far too large a grammar for that, and a comptime parser can't call a C library. The assembler parses it in native code and emits Zig, the same move it already makes everywhere:

```
constants/*.yaml → assembler (native, any YAML lib) → generated constants.zig → game imports
```

Three consequences:

[...]

Issue #809, opened 2026-08-07
RFC: Internationalisation — locale files with comptime-checked keys


apotema:
**RFC:** `RFC-I18N.md` (in this repo, alongside `RFC-FONT-LOADER.md` / `RFC-ASSET-STREAMING.md`)

## Problem

Games hardcode every user-visible string as a Zig literal — e.g. `ui.button("New Game", …)` in `flying-platform-labelle/scripts/menu/menu_ui.zig`. There is no way to ship a second language short of editing source. Nothing in core, engine, gui, or cli mentions locales today.

Rails' shape is the reference: one file per locale, nested keys, `t("menu.new_game")`, interpolation. What is **not** worth copying is the failure mode — Rails finds a typo'd key at lookup time and renders `translation missing: en.menu.new_game` to a user, because nothing validates keys until then.

The assembler already generates the game's wiring at build time. So that entire class of bug can be a **compile error** instead. That is the reason to build this into labelle rather than let each game roll a string table.

## Proposal

**JSONC, not YAML.** labelle already parses JSONC for scenes and prefabs. YAML would add a second config dialect and a parser dependency for one feature. The transferable idea is one-file-per-locale with nested keys, not the syntax.

**Convention-scanned `locales/`**, flat, filename = BCP-47 tag (`en.jsonc`, `pt-BR.jsonc`). No registry to keep in sync — same rule `scripts/` uses. Packs ship `packs/<name>/locales/` with `<pack>__` namespacing, as they already do for components and prefabs.

**Generated nested key constants**, so the dotted call site survives without flattening collisions (`a.b_c` vs `a_b.c`):

```zig
t(K.menu.new_game)   // K.menu.new_gme → compile error at the call site
```

**Usage-aware build-time validation** by the assembler:

| Condition | Result |
|---|---|
| Key **used in code**, in reference, missing from locale *L* | **Warning**, naming the key and the locales missing it |
| Key **unused**, missing from locale *L* | Silent (only under `--verbose`) |
| Key in *L*, absent from reference | Build error — catches half-finished renames |
| Placeholder set differs across locales | Build error, showing both sets |
| `locales/` exists, no `i18n.default` declared | Build error |
| `i18n.default` / `.reference` names a locale with no file | Build error, with "did you mean" |

`i18n.strict = true` promotes row 1 to a hard error — what a release build turns on. Default is warn: a key only the reference locale has yet is the normal state mid-feature.

### Why usage-awareness works here and not in Rails

[...]

PR #811, opened 2026-08-07
docs(rfc): i18n and game constants

apotema:
Two sibling RFCs, no code. Tracked by #809 (i18n) and #810 (constants).

## `RFC-I18N.md` (#809)

Locale files with **comptime-checked keys**. Borrows Rails' shape — one file per locale, nested keys, interpolation — but not YAML, since labelle already parses JSONC for scenes and prefabs.

The part Rails structurally cannot do: its keys are runtime strings and may be computed (`t("menu.#{action}_label")`), so static usage analysis is unsound and every check defers to lookup, rendering `translation missing: …` at a player. labelle's key space is closed and syntactically visible — a `Key` is only nameable as a `K.<path>` constant — which makes coverage decidable at build time.

So diagnostics are **usage-aware**: a key used in code and missing from a locale warns; an unused one is silent. `i18n.strict` promotes it to a comptime error for release builds. Comptime enforces, the assembler advises — Zig has no comptime warning, and `@compileLog` is not one (it ends the build with `error: found compile log statement`, verified on 0.16.0).

`project.labelle` must declare `.default` — no implicit `"en"`. `.reference` (authoring locale, backfill source for gaps) is separate and defaults to it.

## `RFC-CONSTANTS.md` (#810)

`constants/` at project root, one YAML file per domain. Motivated by concrete drift in `flying-platform-labelle`: **38 tuning constants across 17 files**, five construction timings spread over four files with four sharing the value `5.0` and nothing recording whether that is intent, `SHIP_SPEED` defined twice with different values, and decay rates in a plugin's `config.zig` holding the `health_drain_rate = 0.0` that the game's CLAUDE.md spends four paragraphs warning about.

Comptime-only, but **not** via `@embedFile` + comptime parse — YAML is far too large a grammar and a comptime parser cannot call a C library. The assembler parses natively and emits Zig, so the YAML dependency is the assembler's rather than the game's, and values are emitted untyped and coerce at the use site.

Strict scalar policy, because YAML's implicit typing fails silently in exactly this kind of file (`enabled: no` → `false`, `delay: 12:30` → `750`, `id: 0755` → `493`).

## Why together

Both are the same shape: *scan a convention-named folder → generate nested namespaces of comptime-checked accessors → warn on entries nothing uses*. They should share one assembler pass parameterised by root symbol (`K.` / `C.`) rather than growing two near-identical implementations. They differ in one respect only: locales need a runtime-selectable table because the active language changes while the game runs; constants resolve to a single value at build time.

Open questions are listed at the end of each RFC — the load-bearing ones are which YAML library (the assembler has none today), and where the constants/prefab-data boundary sits.

---
<a href="https://app.blacksmith.sh/labelle-toolkit/codesmith/labelle-engine/pr/811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788659975&installation_model_id=427192&pr_number=811&repository=labelle-toolkit%2Flabelle-engine&return_to=https%3A%2F%2Fgithub.com%2Flabelle-toolkit%2Flabelle-engine%2Fpull%2F811&signature=4a0d7498d3e45d6c3262d9db88cad8e7a8bb4eb897178a123bb51815ec7a778d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/RFC-CONSTANTS.md
+++ b/RFC-CONSTANTS.md
@@ -29,7 +29,17 @@ Decay rates cluster in `libs/needs_machine/src/config.zig`, which is the closest
 
 1. **One folder.** `constants/` at project root; every tuning value in the game reachable from one place.
 2. **One file per domain**, named for the domain — `construction.yaml`, `decay.yaml` — so related values sit together and a diff reads as a balance change.
-3. **YAML**, for hand-editing by whoever is balancing the game.
+3. **YAML**, for hand-editing by whoever is balancing the game. This is a
+   deliberate divergence from [RFC-I18N](./RFC-I18N.md) §1, whose argument —
+   "a second config dialect and a new parser dependency for one feature" —
+   applies here too and is overridden, decided, for two reasons. A balance
+   file is the one place the toolkit's files are edited by someone tuning
+   numbers rather than writing code, and `rate: 0.02` with no quotes or
+   braces is the shape that audience edits without breaking. And the parser
+   is not a dependency: it is a hand-written strict subset (§2, Open
+   Question 1 — resolved), small enough that the "new parser" cost is a few
+   hundred lines the assembler owns. The remaining cost is honest: §6's
+   shared pass runs two parsers, one per dialect.
 4. **Zero runtime cost.** Values inline exactly as the `const` literals they replace. A shipped game reads no files and carries no parser.
 5. **A misspelled constant does not compile**, the same property [RFC-I18N](./RFC-I18N.md) §3 gives keys.
 6. **Packs bring their own**, namespaced, so a pack stays drop-in.
@@ -180,7 +190,7 @@ The door is not closed, and it is worth recording the shape now so the decision 
 
 Mirroring [RFC-I18N](./RFC-I18N.md) §3.1: the assembler scans game and pack sources for `C.<path>` references and **warns on constants nothing reads**. A tuning value no code consumes is either dead or a symptom of a rename that missed a call site, and in a folder whose purpose is discoverability, stale entries are the main way it would rot.
 
-Sound for the same reason it is sound there — the accessor path is the only way to name a constant, and it cannot be assembled from a runtime string.
+Sound for the same reason it is sound there — the accessor path is the only way to name a constant, and it cannot be assembled from a runtime string — **with one caveat the scanner must handle: namespace aliasing.** `const cfg = C.decay.hunger;` followed by `cfg.rate` is ordinary Zig, and a textual scan for `C.<path>` misses it, producing a false "unused" warning that trains people to ignore the lint. Ruled: **conservative alias widening** — the scanner detects `= C`-rooted aliases and marks the aliased subtree used. Sound (no warning is wrongly suppressed, no false positive fires); the cost is precision, since an aliased subtree's genuinely dead entries go unreported. The same ruling covers RFC-I18N's `K` (its §3.1).
 
 ### 6. Shared machinery with RFC-I18N
 
@@ -205,7 +215,7 @@ Phase 4 is deliberately last and deliberately incremental. New tuning values go 
 
 ## Open questions
 
-1. **Which YAML library?** The assembler has no YAML dependency today (checked). A pure-Zig parser keeps the toolchain dependency-free and cross-compiles trivially; libyaml through C is more battle-tested but adds a C dependency to a build tool. The strict subset in §2 is small enough that a hand-written parser is also credible — and would make the subset enforceable by construction rather than by post-validation.
+1. ~~**Which YAML library?**~~ **Resolved** (assembler#656): none. The parser is hand-written against the strict subset, which makes §2 enforceable by construction rather than by post-validation — nothing implements YAML 1.1 implicit typing, so nothing can silently apply it. The toolchain stays dependency-free and the subset's rejections are errors that name what was written.
 2. **Do constants and prefab data overlap?** `build_time` as a global vs. a per-workstation override in `prefabs/*.jsonc`. Declared a non-goal here, but the boundary will be contested the first time a designer wants one fast-building room.
 3. **Sequences.** Rejected in phase 1. But tiered values (`upgrade_costs: [10, 25, 60]`) are a natural fit and would generate a comptime array. Worth it, or does it invite structure that belongs in prefabs?
 4. **Units.** `build_time: 5.0` is seconds by convention and comment only. A suffix convention (`_s`, `_px`, `_per_s`) is nearly free and self-documenting; a real unit type is not. Is the convention worth mandating?
--- a/RFC-CONSTANTS.md
+++ b/RFC-CONSTANTS.md
@@ -108,6 +108,19 @@ simpler than i18n's — no "unless it is a new locale" branch — and it is what
 turns a pack renaming a constant in v2 into a build failure rather than a game's
 tuning silently reverting to the pack's default.
 
+In file form the rule follows from §1 — the filename is the namespace, and a
+pack's namespace is `<pack>__<file>` — so the game overrides by carrying the
+prefixed name as a filename:
+
+```yaml
+# constants/citizens__hunger.yaml — the game retuning the citizens pack
+drain_rate: 0.015
+```
+
+The `__` filename is deliberate, not pretty. It keeps "filename is the
+namespace" a single rule with no override-specific carve-out, and a directory
+listing of `constants/` then shows exactly which packs this game has retuned.
+
 An override must also keep the **scalar kind** it replaces: `5.0` stays `5.0`,
 not `5`. Values are emitted untyped and coerce at the use site (§3), so an
 int-for-float override can change behaviour without changing the number — the
--- a/RFC-CONSTANTS.md
+++ b/RFC-CONSTANTS.md
@@ -1,0 +1,162 @@
+# RFC: Game constants — one folder, YAML per domain, inlined at build time
+
+**Status:** Draft
+**Author:** Alexandre
+**Date:** 2026-08-06
+**Sibling RFC:** [RFC-I18N](./RFC-I18N.md) — shares the scan-and-codegen machinery (§6)
+
+## Problem
+
+Tuning values live wherever the code that reads them happens to live. In `flying-platform-labelle` today there are **38 such constants across 17 files**, and the pattern is not that they are hard to find — it is that nothing relates them to each other.
+
+The construction family:
+
+```
+packs/rooms/…/26_construction_progress.zig      BUILD_TIME              = 5.0
+packs/rooms/…/32_carcase_build_dispatcher.zig   CONSTRUCTION_DURATION_S = 5.0
+packs/rooms/…/40_repair_progress.zig            REPAIR_TIME             = 5.0
+packs/rooms/…/40_repair_progress.zig            PER_PIECE_TIME          = 5.0
+packs/rooms/…/29_deconstruction_progress.zig    DECONSTRUCT_TIME        = 4.0
+```
+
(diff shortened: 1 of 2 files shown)


apotema:
## Open Question 4 — resolved: the game takes precedence over the pack

**Yes**, a game can override a pack's string without forking it. And the case is
broader than the question asks: a game can also **add locales the pack never
shipped**. A pack ships `en` and `fr`; the game adds `pt` and others. This is the
normal case for a pack the game does not control.

Same rule for constants (#810): game beats pack.

### What follows

**1. The prefix is invisible inside, explicit outside.**

Inside `packs/citizens` you write `hunger.starving` and never see the namespace,
exactly as RFC-packs promises for components. But a game addressing that key from
outside has no other way to name it, so at the override site it is
`citizens__hunger.starving`. Worth stating, because the packs RFC currently reads
as "authors never type or see the prefix" without qualification.

**2. `.reference` has to become per-pack, or §3.1's guarantee breaks.**

This is the one that actually bites. §3.1 fills every missing slot from the
reference locale so the table is rectangular and *no runtime fallback code needs
to exist*. That assumes the reference locale contains every used key.

For a pack's keys it does not. Pack ships `en` + `fr`; a studio authoring in
Portuguese sets `.reference = "pt"`; the pack's keys exist in neither the game's
`pt` nor anywhere the project reference can see. Nothing to backfill from, and
the table cannot be made rectangular.

Proposal: **a pack declares its own reference in `pack.labelle`**, and backfill
resolves

```
locale → pack reference → project reference
```

A `pt` player then sees the pack's English for strings nobody has translated yet,
which is the right outcome, and "no runtime path can fail" survives contact with
third-party packs.

**3. An override naming a key that does not exist must be an error.**

§3.1 makes unused keys silent. That is right for pack-authored keys and wrong for
overrides. If a game writes `citizens__hunger.starvng`, or the pack renames a key
in v2, the game silently gets the pack's default while believing it retuned
something.

Overriding is an **assertion that the key exists**. It is also the only mechanism
that makes a pack upgrade breaking a game's translations show up at build time
rather than in front of a player.

**4. Adding is not overriding, and only one needs that check.**

A `pt` file for a pack shipping `en`/`fr` is new-locale-for-existing-key. The
check is "does this key exist in the pack's key space", not "does this locale
already have it".

**5. Placeholders come free from §4 — as long as overrides go through the same
codegen.** An added `pt` string that drops `{count}` should fail exactly like a
game's own would. Cheap, and the alternative fails at a player.

### For #810 (constants)

The precedence rule is the same, but constants need something translations do
not: **which constants are overridable is a pack's public API.** Packs already
have `.global` vs `.pack` visibility for components, defaulting to private, so a
pack can refactor without a game reaching into its internals. Without the same
split for constants, every constant becomes load-bearing the moment one game
overrides it and a pack can never rename an internal value again. All strings are
legitimately overridable; not all constants are.

Also: an override should keep the scalar kind (`5.0`, not `5`). Values are emitted
untyped and coerce at the use site, so an int-for-float override can change
behaviour without changing the number — which is the same class of silent failure
the strict scalar policy exists to prevent.

apotema:
## Correction to the previous comment — constants work the same way

Decided: **constants follow the same rule as translations.** A game can override
any of a pack's constants, with the game taking precedence. No separate
visibility gate.

That supersedes the `.global` / `.pack` split I floated in the "For #810" section
above. Recording what it means so it is a decision rather than an omission:

**A pack's constant names become part of its compatibility surface.** Once a game
overrides `citizens.hunger.drain_rate`, renaming it in the pack is a breaking
change for that game. This is the same bargain packs already accept elsewhere —
RFC-packs makes a pack's `name` save-stable because it is the component prefix,
so renaming a shipped pack is a save migration. Constant names join that list.

It is the right trade for the problem #810 actually documents. The motivating
evidence is 38 tuning constants across 17 files, `SHIP_SPEED` defined twice with
different values, and `health_drain_rate = 0.0` sitting in a plugin's
`config.zig`. A visibility gate would mean a pack author has to predict which
values a game will want to tune, and every wrong guess sends that game back to
forking or to redefining the value somewhere else — which is exactly the drift
being fixed.

### One asymmetry worth naming

Translations have two operations; constants have one.

For translations, *adding* is a real case — a `pt` file for a pack that ships
`en`/`fr` is new-locale-for-existing-key, and the check is "does this key exist in
the pack's key space".

Constants have no second axis. There is no "add a constant to someone else's
pack": a value the pack does not define is the game's own constant, in the game's
own namespace. So for constants, **every write under a pack's namespace is an
override**, and rule 3 applies without exception — one that matches no existing
constant is an error, always. No adding case to carve out.

That makes the constants rule strictly simpler than the i18n one, which is a
point in favour of the shared assembler pass the RFC proposes: same scan, same
precedence, same must-exist check, and only i18n needs the extra "or it is a new
locale" branch.

apotema:
Folded both decisions into the RFCs — `ffab6f9`.

**RFC-I18N** gains §2.1 (the game overrides *and* adds; the namespace is
invisible inside a pack and explicit outside; writing under a pack's namespace
asserts the key exists) and §2.2 (a pack declares its own `.reference`, backfill
resolves `locale → pack reference → project reference`). Open Question 4 is
struck through and marked resolved, and the §3.1 backfill sentence now names
which reference applies to a `<pack>__` key.

**RFC-CONSTANTS** gains §1.1 — same precedence, no visibility gate, every write
under a pack's namespace is an override, scalar kind must be preserved — and the
accepted consequence is written down rather than left implicit: a pack's constant
names become part of its compatibility surface.

§6 also changed. The two RFCs now differ in *two* respects rather than one, and
the second argues **for** the shared pass rather than against it: constants have
no counterpart to i18n's adding case, so the constants rule is a strict subset of
the i18n one. Build the general form once, let constants use the simpler half.

Phasing rows for both updated — this work belongs with pack-scoped scanning,
which is already phase 3 in each.

apotema:
## Evaluation

Full read of both documents. **The direction is right and the core claim survives scrutiny** — comptime-checked keys over assembler codegen is a real differentiator, and the honesty about Zig's limits (no comptime warning; `@compileLog` behaviour verified) is what makes the argument trustworthy. Both problem statements argue from named files and real values, which is the strongest part of the pair.

Five findings. **2, 4 and 5 are already fixed** (`ed99d09`, plus `e0b09a3` for two defects my own earlier edit introduced — a wrong cross-reference and a stranded example). **1 and 3 need a ruling.**

### 1. OPEN — the usage scanner is unsound under namespace aliasing (both RFCs)

§3.1(b) rests on "a `Key` is only nameable as a `K.<path>` constant". The stated hole is `@enumFromInt` — pathological, fairly out of contract. The unstated one is ordinary Zig:

```zig
const S = K.settings;     // any settings screen will write this
ui.label(t(S.volume));
```

A textual scan for `K.<path>` misses it → `S.volume` is classified *unused* → its missing-translation warning is **suppressed** → a hole a player can see ships silently. For constants (§5) the same aliasing produces false "unused constant" warnings, which trains people to ignore the lint — the exact rot the section exists to prevent. The strict/comptime path (a) is unaffected.

Options, cheapest first: (i) declare aliasing of `K`/`C` subtrees out of contract — weak, it is too natural to ban; (ii) detect `= K`/`= C` aliases and conservatively mark the aliased subtree used — loses precision, keeps soundness, probably the right first cut; (iii) AST-aware scanning — right, costs more. The RFC should pick one; as written it claims a soundness the scanner does not have.

### 2. FIXED — interpolation was unspecified for the case that is i18n's whole point (I18N §4)

Placeholder *set* parity was checked; placeholder *order* legitimately differs (`"{count} of {max}"` vs `"Von {max} Artikeln: {count}"`). The string is chosen at runtime, the args at comptime — a naïve comptime-format-string implementation renders every reordering locale wrong, and nothing forbade it. Now specified: codegen emits per-(key, locale) segment lists, `tf` walks the active one, parity checks sets never order. `{{` escapes a brace.

### 3. OPEN — the two RFCs contradict each other on format, and only one argues its side

I18N §1 rejects YAML: *"a second config dialect and a new parser dependency for one feature."* CONSTANTS then chooses YAML, justified only as "for hand-editing" — which JSONC satisfies too. The choice's cost is visible in the document itself: **§2 exists entirely to defuse YAML's implicit-typing hazards, Open Question 1 exists to pick a YAML library, and §6's "one shared pass" quietly needs two parsers.** Choosing JSONC deletes §2 and OQ1 and completes §6; the loss is unquoted keys/values in flat balance files. Either is defensible — but CONSTANTS §1 must defend YAML *against its sibling's own argument*, which it currently does not acknowledge.

### 4. FIXED — the override authoring syntax was never shown (both, phase 3)

The semantics from the precedence decisions were precise; no file was ever shown. Now: a pack's keys sit at the top level of the game's locale file under the prefixed name (`"citizens__hunger": { … }` in `locales/pt.jsonc`), and constants keep "filename is the namespace" whole — the game overrides `C.citizens__hunger.*` with `constants/citizens__hunger.yaml`. Recorded as deliberate-not-pretty: a listing of `constants/` shows exactly which packs the game has retuned.

### 5. FIXED — the prior-art claim was false, and the counterexample helps (I18N)

*"None of the prior art can check at compile time"* — Android resources falsify it: `R.string.foo` is a generated constant (missing key = compile error), untranslated strings are lint warnings, default-locale fallback is baked. That is this RFC's exact two-severity split, shipped at scale, so citing it turns the design from speculative into proven. The still-novel part is sharpened to what survives: **usage-awareness**. Android's lint must assume any string might be used because `getIdentifier()` resolves from runtime strings; labelle's closed key space is what makes "is this key actually rendered" decidable.

### Smaller notes, no fix applied

- **§2.2 semantic drift**: a game overriding a pack key in the pack's *reference* locale changes the backfill source, but the pack's existing `fr`/`de` still translate the original meaning. Uncatchable by any check; deserves one acknowledging sentence.
- **§3's validation table predates packs**: "absent from reference → build error" needs "reference" to mean the owning realm's reference (§2.2), or the game-adds-`pt` case trips it.
- **OQ3 now interacts with phase 3**: index stability must span the game/pack merge.
- **CONSTANTS §1.1's scalar-kind rule** would land harder with its concrete failure: `const x = C.a.b; x / 2` — integer vs float division under inference.

### Not to be touched

The evidence-first problem statements; §3.1's open-vs-closed key space contrast (the sharpest paragraph in either doc); the rectangular-table/no-runtime-fallback guarantee, which the per-pack `.reference` decision preserved; the `t`/`tf` split; §6 refusing an interface slot for one implementation; CONSTANTS' incremental phase-4 migration; hot-reload's door-open shape.

apotema:
Implementation has started: **labelle-toolkit/labelle-assembler#656** is CONSTANTS phase 1 — the strict-subset parser (hand-written, resolving Open Question 1), the scan-and-emit phase, and the build wiring. Verified end to end on `examples/null`.

Three rulings folded into the RFCs in `1cbfe2d`:
- **constants first**, as the simpler half of the shared pass
- **YAML stays**, now defended against RFC-I18N §1's own argument rather than contradicting it silently; OQ1 resolved (no library — the subset is enforced by construction)
- **conservative alias widening** for the usage scanner, recorded in both documents (I18N §3.1, CONSTANTS §5) — `const cfg = C.decay.hunger;` is ordinary Zig a textual scan misses, and widening is the sound answer

